PR #2, opened 2020-02-20
XLM-R model as encoder

Mehrad0711:


Pipfile
```diff
@@ -16,6 +16,7 @@ tensorboardX = "==2.0.*"
 requests = "~=2.22"
 transformers = "~=2.3"
 radam = {git = "https://github.com/LiyuanLucasLiu/RAdam"}
+sentencepiece = ">=0.1.83"
```
gcampax: you need to add the dependency to setup.py too, also you should probably set the dependency to be "<0.2.0" as well.

genienlp/predict.py
```diff
@@ -36,7 +36,7 @@
 import torch
 
 from . import models
-from .data.embeddings import load_embeddings
+from .data_utils.embeddings import load_embeddings
```
gcampax, line 39: Why so many file renames?

genienlp/tasks/base_dataset.py
```diff
@@ -1,11 +1,11 @@
 import os
 import zipfile
 import tarfile
+from six.moves import urllib
```
gcampax: While we're changing this, we should import the python3 version directly, instead of importing from six. We don't want six.

gcampax (approved):
Looks good! Lots of unrelated changes (sad!) but ok. Please fix the dependency in setup.py then push.

gcampax:
Also see if you can fix the LGTM stuff while you're here.

Pipfile
Mehrad0711: fixed

genienlp/predict.py
Mehrad0711: it's more intuitive (e.g. data folder does not contain the data itself) Also easier to infer dependencies between files in the same package.

genienlp/tasks/base_dataset.py
Mehrad0711: fixed

Mehrad0711:
I made the changes and fixed the lgtm alerts as well.

gcampax:
Ok yeah please merge

PR #4, opened 2020-03-02
Bug fixes and updates

Mehrad0711:
- Fixing XLM-R decoding 
- Make multi-gpu training work (need to average the losses returned by GPUs)
- Fixing the loading problem when using previous checkpoints to resume training


gcampax (approved):
Ya

PR #6, opened 2020-03-16
First batch of changes from dialogue work

gcampax:
- Introduces the new Almond tasks for dialogue
- Modifies the Coattention and BiLSTM models
- Separates context and question embeddings
- Adds an option to pretrain the context embedding with MLM
- Adds the option to only fine-tune the embedding after a certain number of iterations


gcampax:
I'm just going to merge this, after we discussed on Slack.

PR #10, opened 2020-03-25
Wip/multilanguage

Mehrad0711:
PR for multilingual project:

- Almond(and other tasks later) can have multiple train/dev/test sets all under its root path for each language
- Allow combined or separate validation on each dev data
- Allow caching multiple transformer embeddings using '+'
- Using namedTuple for splits preventing possible bugs


Mehrad0711:
ToDo: update test cases

genienlp/tasks/almond/__init__.py
```diff
@@ -31,86 +31,83 @@
...
     base_url = None
 
-    def __init__(self, path, *, make_example, subsample=None, cached_path=None, skip_cache=False,
-                 **kwargs):
+
+    def __init__(self, path, *, make_example, subsample=None, cached_path=None, skip_cache=False, examples=None, **kwargs):
+        
+        if examples:
+            super().__init__(examples, **kwargs)
+            return
```
gcampax: What's this code about? Where are you calling AlmondDataset with the example list? This looks very ugly.

genienlp/tasks/almond/__init__.py
```diff
@@ -312,4 +309,54 @@ def _make_example(self, parts):
                                 tokenize=self.tokenize, lower=False)
 
     def get_splits(self, root, **kwargs):
-        return AlmondDataset.splits(root=root, task_name='almond/agent', make_example=self._make_example, **kwargs)
\ No newline at end of file
+        return AlmondDataset.return_splits(path=os.path.join(root, 'almond/agent'), make_example=self._make_example, **kwargs)
+    
+    
+@register_task('almond_multilingual')
+class AlmondDialoguePolicy(BaseAlmondTask):
```
gcampax: class name here

genienlp/tasks/almond/__init__.py
```diff
@@ -312,4 +309,54 @@ def _make_example(self, parts):
...
+    def combine_datasets(self, datasets):
+        splits = defaultdict()
+        assert len(datasets) >= 1
+        fields = [field for field in datasets[0]._fields if getattr(datasets[0], field) is not None]
+        
+        for field in fields:
+            all_examples = []
+            for dataset in datasets:
+                all_examples.extend(getattr(dataset, field).examples)
+            splits[field] = AlmondDataset('', make_example=None, examples=all_examples)
```
gcampax: it doesn't need to be an AlmondDataset here, right? it could be a plain dataset

genienlp/tasks/almond/__init__.py
```diff
@@ -31,86 +31,83 @@
...
             os.makedirs(os.path.dirname(cache_name), exist_ok=True)
             logger.info(f'Caching data to {cache_name}')
             torch.save(examples, cache_name)
 
         super().__init__(examples, **kwargs)
+        
 
     @classmethod
-    def splits(cls, root='.data', task_name='almond', train='train', validation='eval', test='test', **kwargs):
+    def return_splits(cls, path, train='train', validation='eval', test='test', **kwargs):
```
gcampax, line 80: Why are you renaming this method and not the one in the other classes?

genienlp/train.py
```diff
@@ -76,40 +76,37 @@ def prepare_data(args, logger):
...
+        kwargs.update({'subsample': args.subsample, 'skip_cache': args.skip_cache,
+                       'cached_path': os.path.join(args.cache, task.name), 'languages': args.train_languages})
         if args.use_curriculum:
             kwargs['curriculum'] = True
-        kwargs['skip_cache'] = args.skip_cache
-        kwargs['cached_path'] = os.path.join(args.cache, task.name)
 
         logger.info(f'Adding {task.name} to training datasets')
         split = task.get_splits(args.data, lower=args.lower, **kwargs)
+        assert not split.eval and not split.test
```
gcampax, line 87: This is a very weird way to write this code, basically you have an API that loads 3 datasets at once (train dev test) but you never use it that way, only ever load one dataset at a time.

gcampax (approved):
It's a big change very close to a release, but it looks ok. Couple style comments really.
Please get the tests to pass then land it quickly. If it's not by today or tomorrow at the latest, it will not land until after the release (Apr 3)

genienlp/train.py
Mehrad0711: yeah, it is weird, but it only loads the split if it supposed to. Otherwise, it will return None instantly.

genienlp/tasks/almond/__init__.py
Mehrad0711: because I am using 'path' instead of 'root' and pylint complains that method does not match with Dataset class.

genienlp/tasks/almond/__init__.py
Mehrad0711: True. I wanted to be consistent so each task uses its own dataset. I can change it though.

genienlp/tasks/almond/__init__.py
Mehrad0711: it is called here:
`splits[field] = AlmondDataset('', make_example=None, examples=all_examples)`
getting rid of it depends on the resolution of your next comment.

genienlp/tasks/almond/__init__.py
gcampax: As the other comment says, it's very ugly to have two completely different signatures for the same function, without sharing any of the code.

Mehrad0711:
Ok. the PR is ready for final review and merge.

PR #12, opened 2020-03-31
Fix server for multilingual_eval

gcampax:
@Mehrad0711 you mentioned a bug in predict.py I wonder if it's the same?
This is for server.py though


Mehrad0711:
no the splits for prediction are switched. I tagged you in my PR.

PR #14, opened 2020-04-06
Wip/features

Mehrad0711:


gcampax:
How is this meant to be used? Is it a multilingual Almond-only thing?

Mehrad0711:
It can be used for all tasks, but yeah I'm using for multilingual.
It is useful for cases when train split is the same but you have multiple dev sets.

Also, it's backward compatible with the previous release so it can be merged.

gcampax:
I still don't understand how this works. What strings can you pass? Is it just "eval" or "aux", or something else? What about non-Almond tasks, which don't call their dev set eval.tsv ?

Mehrad0711:
It should be the name of the file you want to use (without the file extension) for evaluation.
For example you can choose from 'eval-fa' and 'eval-en' and 'eval-it' which are separate tsv files in the almond data directory. 
You are right about other tasks. I removed the default value so now it's only effective if the argument is provided.

gcampax:
It still doesn't make sense to me. Other tasks might not even have multiple files for the different sets.

If you have multiple dev sets, just copy/symlink the file in the right place before you call genienlp. This is a no for me.

PR #15, opened 2020-04-14
Wip/multilanguage

Mehrad0711:
Adding option to have paired sentences as context input (pairs should be semantically-equivalent sentences: translations or paraphrases of each other)

### DIFF
--- a/genienlp/tasks/generic_dataset.py
+++ b/genienlp/tasks/generic_dataset.py
@@ -53,15 +53,9 @@ def make_example_id(dataset, example_id):
 def context_answer_len(ex):
     return interleave_keys(len(ex.context), len(ex.answer))
 
-def processed_id(ex):
+def id_value(ex):
     id_ = ex.example_id.rsplit('/', 1)
     id_ = id_[0] if len(id_) == 1 else id_[1]
-    # translated
-    if id_[0] == 'T':
-        id_ = id_[1:]
-    # paraphrased
-    if id_[0] == 'P':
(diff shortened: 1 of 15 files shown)


genienlp/tasks/generic_dataset.py
```diff
@@ -53,11 +53,15 @@ def make_example_id(dataset, example_id):
...
-def same_id(ex):
+def processed_id(ex):
     id_ = ex.example_id.rsplit('/', 1)
     id_ = id_[0] if len(id_) == 1 else id_[1]
+    # translated
     if id_[0] == 'T':
         id_ = id_[1:]
+    # paraphrased
+    if id_[0] == 'P':
+        id_ = id_[1:]
```
gcampax: ??? P means PPDB, not paraphrase. Also, why do you care? Stop putting these hacks in genienlp. They don't belong here.

gcampax (changes requested):
More hacks. More options. More "lets add stuff to the library and see what sticks". I'm not happy with this. I'm 100% sure we'll cut out 90% of what you're adding, before the end of the quarter. Just keep this stuff in your branch, alright.

genienlp/tasks/generic_dataset.py
Mehrad0711: fixed.

Mehrad0711:
[resolved] showing improvement over baseline for translation (2%) and paraphrase (5%) training tasks.

PR #16, opened 2020-04-19
Wip/paraphrase

s-jse:


genienlp/data_manipulation_scripts/clean_paraphrasing_dataset.py
```diff
@@ -32,31 +35,53 @@ def main():
...
 
     args = parser.parse_args()
 
     drop_count = 0
     # number_of_lines = get_number_of_lines(args.input)
+    # number_of_lines = get_number_of_lines(args.input)
+    output_size = 0
     with open(args.input, 'r') as input_file, open(args.output, 'w') as output_file:
+        writer = csv.writer(output_file, delimiter='\t')
         reader = csv.reader(input_file, delimiter='\t')
```
gcampax, line 58: Actually, you cannot use the CSV module to read our datasets, because they are not proper TSVs: fields are not delimited with `"`, and `"` is not escaped when it appears in a field. You should should split on `\t` directly.

genienlp/data_manipulation_scripts/dialog_to_tsv.py
```diff
@@ -0,0 +1,92 @@
...
+    parser.add_argument('--agent_prefix', type=str, default='A:', help='')
+    parser.add_argument('--at_prefix', type=str, default='AT:', help='')
+    parser.add_argument('--ut_prefix', type=str, default='UT:', help='')
+
+    args = parser.parse_args()
+
+    all_prompts = read_dialog_file(args.dialog_file, args)
+    
+    with open(args.input) as input:
+        reader = csv.reader(input, delimiter='\t')
```
gcampax, line 75: Same here, you cannot use the CSV module to read from a Genie TSV dataset

genienlp/data_manipulation_scripts/dialog_to_tsv.py
```diff
@@ -0,0 +1,92 @@
...
+                previous_ats = []
+                next_uts = []
+        if l.startswith(args.agent_prefix):
+            agent_utterances.append(l[len(args.agent_prefix):].strip())
+        if l.startswith(args.user_prefix):
+            user_utterance = l[len(args.user_prefix):].strip()
+            last_agent_utterance = agent_utterances[-1]
+            prompts.append((last_agent_utterance, user_utterance))
+
+    return prompts
```
gcampax, line 53: This whole file and this code would be dramatically easier if written in JavaScript, because Genie already has code to parse its own dialogue format, which also appears to be more robust to invalid inputs than this.

genienlp/run_generation.py
```diff
@@ -201,76 +231,208 @@ def sample_sequence(model, length, context, position_ids, num_samples=1, tempera
...
+    SpecialTokenMap('EMAIL_ADDRESS_([0-9]+)', ['e1@example.com', 'e2@example.com']),
+    SpecialTokenMap('URL_([0-9]+)', ['my1site.com', 'my2site.com']),
+    SpecialTokenMap('DATE_([0-9]+)', ['5-6-2015', '8-3-2016']),
+    SpecialTokenMap('CURRENCY_([0-9]+)', ['$12', '$13'], [['$12', 'twelve dollars', '12 dollars', '$ 12', '$ 12.00', '12.00', '12'], 
+                                                          ['$13', 'thirteen dollars', '13 dollars', '$ 13', '$ 13.00', '13.00', '13']]),
+    SpecialTokenMap('DURATION_([0-9]+)', ['5 weeks', '6 weeks'], [['5 weeks', 'five weeks'], ['6 weeks', 'six weeks']]),
+    SpecialTokenMap('LOCATION_([0-9]+)', ['locatio1n', 'locatio2n'], [['locatio1n', 'locat1n'], ['locatio2n', 'locat2n']]),
+    SpecialTokenMap('QUOTED_STRING_([0-9]+)', lambda x: 'Chinese', lambda x: ['Chinese', 'chinese']), # TODO change to be more general than cuisine
+    SpecialTokenMap('GENERIC_ENTITY_uk.ac.cam.multiwoz.Restaurant:Restaurant_([0-9]+)', ["restaurant1", "restaurant2", "restaurant3"]) # TODO the only reason we can get away with this unnatural replacement is that actual backward is not going to be called for this
```
gcampax, line 258: Do you have a plan for the TODOs here? I feel like we'll soon forget about this.

genienlp/run_generation.py
```diff
@@ -201,76 +231,208 @@ def sample_sequence(model, length, context, position_ids, num_samples=1, tempera
...
+                    sentences[idx] = sentences[idx].replace(p, capitalized_p)
+            sentences[idx] = sentences[idx].replace(' i ', ' I ')
+            sentences[idx] = sentences[idx][0].upper()+sentences[idx][1:]
+            
+    s = ' '.join(sentences)
+    if not keep_tokenized:
+        s = detokenize(s)
+    
+    if not is_cased:
+        s = lower_case(s)
```
gcampax, line 367: This file is _definitely_ not the place for this kind of (language-specific) heuristics. Either move them to a preprocessing of the dataset, or somewhere else.

genienlp/run_generation.py
```diff
@@ -279,37 +441,75 @@ def parse_argv(parser):
     parser.add_argument("--input_file", type=str, help="The file from which we read prompts.")
     parser.add_argument('--input_column', type=int, required=True,
                         help='The column in the input file which contains the input sentences.')
+    parser.add_argument('--prompt_column', type=int, default=None,
+                        help='The column in the input file which contains the text we should start generation from.')
+    parser.add_argument('--gold_column', type=int, default=None,
+                        help='The column in the input file which contains the gold sentences. Defaults to --input_column if no gold is available.')
+    parser.add_argument('--thingtalk_column', type=int, default=None,
+                        help='The column in the input file which contains the ThingTalk program.')
```
gcampax, line 450: Why do you need the ThingTalk program in this script? I thought we do the filtering as a separate pass?

genienlp/run_generation.py
```diff
@@ -379,101 +577,130 @@ def run_generation(args):
...
+            all_outputs.extend(batch_outputs)
+        else:
+            for idx, example in enumerate(batch_outputs):
+                print('input sequence: ', batch_input_sequences[idx % batch_size])
+                c, example = tuple(zip(*sorted(list(zip(batch_criterion[idx], example)), reverse=True)))
+                print(example)
+                print(c)
+                print('-'*10)
+                selection = example[0]
+                all_outputs.append([selection])
```
gcampax, line 692: This whole function needs to be split into multiple functions, it's already way too long and too nested to read.

genienlp/train.py
```diff
@@ -165,7 +165,7 @@ def train_step(model, batch, iteration, opt, devices, lr_scheduler=None, grad_cl
         opt.zero_grad()
     loss, predictions = model(batch, iteration, pretraining=pretraining)
     if torch.isnan(loss).any():
-        raise RuntimeError('Got NaN loss')
+        raise RuntimeError('Got NaN loss %s', str(loss))
```
gcampax, line 168: Wouldn't `str(loss)` be `nan`? Why do you want to print it?

genienlp/train.py
```diff
@@ -481,6 +481,14 @@ def init_model(args, numericalizer, context_embeddings, question_embeddings, dec
     return model
 
 
+def get_linear_schedule_with_warmup(i, dimension, warmup, num_training_steps):
+    i += 1
+    warmup = max(1, warmup)
+    if i < warmup:
+        return float(i) / float(max(1, warmup))
+    return 1. / math.sqrt(dimension) * max(0.0, float(num_training_steps - i) / float(max(1, num_training_steps - warmup)))
```
gcampax: Do we want an option for this? Or is this debugging code / experimental code that is unused?
In the latter case, kill it.

genienlp/util.py
```diff
@@ -45,17 +45,123 @@
...
     for t in tokens:
-        text = text.replace(' ' + t, t)
-    text = text.replace("( ", "(")
-    text = text.replace('gon na', 'gonna')
-    text = text.replace('wan na', 'wanna')
-    return text
+        string = string.replace(' ' + t, t)
+    string = string.replace("( ", "(")
+    string = string.replace('gon na', 'gonna')
+    string = string.replace('wan na', 'wanna')
```
gcampax, line 128: The full (kind of) list of detokenization tokens and assimilations is at https://github.com/stanford-oval/genie-toolkit/blob/master/lib/i18n/american-english.js#L154
Which of course, makes me wonder why we're duplicating this code in python when it could be as well written in JS!

gcampax (changes requested):
A lot of comments, some important ones

genienlp/run_generation.py
s-jse: It is necessary to identify parameter values and capitalize them for cased models.

gcampax: Wouldn't that be a separate preprocessing? It has nothing to do with generation...

genienlp/train.py
s-jse: I don't quite remember how this was a problem, but the difference is that on multiple GPUs, some elements of loss might be NaN, not all of them, and we are using `.any()`.

genienlp/data_manipulation_scripts/clean_paraphrasing_dataset.py
s-jse: With `delimiter='\t'`, csv reader behaves more or less like spitting on `\t`. It has not had any issues with `"`, but it might for special cases. I'll fix this.

PR #17, opened 2020-04-28
Wip/mbart

Mehrad0711:


s-jse:
I have already merged evaluate_bart and run_generation.
There are test files in `tests/dataset/paraphrasing/` that work with run_generation. There is no need for `tests/dataset/paraphrasing/mbart/`.

s-jse:
I think MBART doesn't work yet. I recommend adding it when it starts working.

Mehrad0711:
I think it is better to keep BART training and evaluation separate until we merge my task (sentence correction) and your task (paraphrasing)
Sure I will use the paraphrasing dataset for test.

MBART is not ready yet but BART is. I named the model MBART (rather than BartSystem) to avoid renaming later. 
Main reason for this PR is to avoid code duplication and parallel work as we discussed.

s-jse:
The thing is, run_generation already supports all seq2seq tasks, not just paraphrasing.

Mehrad0711:
yes we will need multiple levels of code merging. run_generation is almost there but still needs bart integration. Then we have to move bart training to run_lm_finetuning. And possibly an eventual merge with the GeneiNLP library cause some of the code is already duplicated.

gcampax:
> The thing is, run_generation already supports all seq2seq tasks, not just paraphrasing.

Yeah I was actually wondering about this: given the new Pipeline code in the transformers library, and the fact that they support seq2seq now, can we ditch most of our training code and switch to more-or-less vanillay transformers? With a couple layers for data loading.
Cause I really don't like having two entirely separate training loops in the repo, and two non-interacting set of scripts. It leads to all sort of confusion, such as this discussion here.

Mehrad0711:
I think it's good to have a chat about "the future" of GenieNLP library!

PR #18, opened 2020-05-12
A potentially better way to fine-tune BERT for contextual tasks

s-jse:
- Using `--append_question_to_context_too` will append a copy of the question to the end of context, and will tokenize it like a non-program field.
This approach seems effective especially for the `almond_dialogue_nlu` task.
- transformers is updated to version 2.9. Due to changes in model hashes, rebuilding the common docker image might be necessary.

### DIFF
--- a/genienlp/arguments.py
+++ b/genienlp/arguments.py
@@ -175,6 +175,8 @@ def parse_argv(parser):
                         help='probability of replacing a token with mask for MLM pretraining')
     parser.add_argument('--force_subword_tokenize', action='store_true', default=False,
                         help='force subword tokenization of code tokens too')
+    parser.add_argument('--append_question_to_context_too', action='store_true', default=False,
+                        help='')
 
     parser.add_argument('--warmup', default=800, type=int, help='warmup for learning rate')
     parser.add_argument('--grad_clip', default=1.0, type=float, help='gradient clipping')
--- a/genienlp/data_utils/example.py
+++ b/genienlp/data_utils/example.py
@@ -45,13 +45,15 @@ class Example(NamedTuple):
     question_word_mask: List[bool]
     answer: List[str]
     answer_word_mask: List[bool]
+    context_plus_question: List[str]
+    context_plus_question_word_mask: List[bool]
 
     vocab_fields = ['context', 'question', 'answer']
 
     @staticmethod
     def from_raw(example_id: str, context: str, question: str, answer: str, tokenize, lower=False):
         args = [example_id]
-        for argname, arg in (('context', context), ('question', question), ('answer', answer)):
+        for argname, arg in (('context', context), ('question', question), ('answer', answer), ('question', context+' '+question)):
             words, mask = tokenize(arg.rstrip('\n'), field_name=argname)
             if mask is None:
                 mask = [True for _ in words]
@@ -59,6 +61,7 @@ class Example(NamedTuple):
                 words = [word.lower() for word in words]
             args.append(words)
             args.append(mask)
+        
         return Example(*args)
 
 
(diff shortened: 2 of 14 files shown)


genienlp/arguments.py
```diff
@@ -175,6 +175,8 @@ def parse_argv(parser):
                         help='probability of replacing a token with mask for MLM pretraining')
     parser.add_argument('--force_subword_tokenize', action='store_true', default=False,
                         help='force subword tokenization of code tokens too')
+    parser.add_argument('--append_question_to_context_too', action='store_true', default=False,
```
gcampax, line 178: The argument name is tad verbose, IMHO

genienlp/data_utils/example.py
```diff
@@ -45,20 +45,23 @@ class Example(NamedTuple):
...
+    context_plus_question_word_mask: List[bool]
 
     vocab_fields = ['context', 'question', 'answer']
 
     @staticmethod
     def from_raw(example_id: str, context: str, question: str, answer: str, tokenize, lower=False):
         args = [example_id]
-        for argname, arg in (('context', context), ('question', question), ('answer', answer)):
+        for argname, arg in (('context', context), ('question', question), ('answer', answer), ('question', context+' '+question)):
             words, mask = tokenize(arg.rstrip('\n'), field_name=argname)
```
gcampax, line 57: I see we pass 'question' here to the Task class, so for Almond tasks everything will be subword split (mask all True). Just confirming that's the expected behavior.

genienlp/tasks/almond/__init__.py
```diff
@@ -233,7 +234,10 @@ class AlmondDialogueNLU(BaseAlmondTask):
     state of the conversation)
     """
     def _is_program_field(self, field_name):
-        return field_name in ('answer', 'context')
+        if self.append_question_to_context_too:
+            return field_name == 'answer'
+        else:
+            return field_name in ('answer', 'context')
```
gcampax: You don't need this, cause you'll get "question" as the field name for the concatenated field?

gcampax (approved):
This looks good. Clean and neat. Looking forward to try it.
Couple comments just to check.

gcampax:
Btw, for the transformers version you need to change setup.py too, because we don't use pipenv in the docker.
(In fact, we should probably get rid of the Pipfile entirely, it's just confusing)

genienlp/tasks/almond/__init__.py
s-jse: Yes, it is a remnant of another approach. I'll remove it.

genienlp/data_utils/example.py
s-jse: Yes, masks will be all True except for words that are entities (start with capital letters).

genienlp/arguments.py
s-jse: This option appends the question to context, but feeds in the original question separately as well, hence "too". What do you suggest?

gcampax: I was going to suggest just "append_question_to_context". Shorter and equally unambiguous.

genienlp/run_generation.py
```diff
@@ -152,8 +152,8 @@ def sample_sequence(model, length, context, position_ids, num_samples=1, tempera
                 inputs['past'] = past
                 if past is not None:
                     inputs['input_ids'] = next_token
-                    inputs['position_ids'] = position_ids[:, next_index-1]
-                    inputs['token_type_ids'] = segment_ids[:, next_index-1]
+                    inputs['position_ids'] = position_ids[:, next_index-1].unsqueeze(-1)
+                    inputs['token_type_ids'] = segment_ids[:, next_index-1].unsqueeze(-1)
```
gcampax, line 156: This was not "Remove unused code". Sneaked in by accident?

s-jse: Should have been two separate commits, but the change was intentional. It failed one of the paraphrasing tests with the newer transformers version, so I had to make this change. All this code will be obsolete when we merge `wip/paraphrase` branch.

PR #19, opened 2020-05-13
Wip/mehrad/fixes

Mehrad0711:
Changes:
- Cache any transformer embedding (so we can use bart and mbart)
- Log context during prediction

### DIFF
--- a/genienlp/data_utils/embeddings.py
+++ b/genienlp/data_utils/embeddings.py
@@ -32,7 +32,7 @@ import torch
 import os
 from collections import defaultdict
 import logging
-from transformers import AutoTokenizer, AutoModel, AutoConfig
+from transformers import AutoTokenizer, AutoModel, AutoConfig, ALL_PRETRAINED_MODEL_ARCHIVE_MAP
 from typing import NamedTuple, List
 
 from .numericalizer.simple import SimpleNumericalizer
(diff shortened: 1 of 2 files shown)


genienlp/data_utils/embeddings.py
```diff
@@ -214,7 +213,7 @@ def load_embeddings(cachedir, context_emb_names, question_emb_names, decoder_emb
             continue
 
         emb_type = get_embedding_type(emb_name)
-        if _is_bert(emb_name) or _is_xlmr(emb_name):
+        if emb_name in ALL_PRETRAINED_MODEL_ARCHIVE_MAP.keys():
```
gcampax, line 216: Don't you need Tokenizer subclasses for new models?

Mehrad0711: So this is meant to be used with "cache-embeddings" command to cache models locally or for the docker. The model and tokenizer will be instantiated from "run_lm_finetune" and "run_generation" script.

gcampax: Yeah then we don't need it, because the docker doesn't cache the pretrained models anymore.

Mehrad0711: Mbart models are pretty big and it's better to cache them (in our own dockers). It speeds up subsequent training.

gcampax: You still have to download the docker image on each machine for each training. The difference is not that big...

Mehrad0711: Agreed

PR #20, opened 2020-05-21
Treat device names as special tokens

s-jse:
So that uncased transformer models do not replace them with [UNK] token.


gcampax (approved):
Sure

PR #22, opened 2020-05-26
Add an option to turn the Almond context into something more English-like

gcampax:
Remove enum:, param: so BERT has more knowledge of what these
words mean.


s-jse (commented):
Looks good.

PR #23, opened 2020-05-31
Add the option to override the question for all tasks

gcampax:
Together with --append_question_to_context, this will append the
true question to the context, and use a fake question as input
to the model. Useful for ablation.


gcampax:
This was used in the EMNLP paper in the end, and it appears to work, so I'm pushing it.

PR #25, opened 2020-06-09
Fix multigpu training

Mehrad0711:


Pipfile
```diff
@@ -11,7 +11,7 @@ scikit-learn = "*"
 
 [packages]
 numpy = "*"
-torch = "~=1.4"
+torch = "==1.4"
```
gcampax: Why this change?

genienlp/train.py
```diff
@@ -162,7 +162,8 @@ def train_step(model, batch, iteration, opt, devices, lr_scheduler=None, grad_cl
                                                iteration > train_question_embeddings_after)
     if (iteration) % gradient_accumulation_steps == 0:
         opt.zero_grad()
-    loss, predictions = model(batch, iteration, pretraining=pretraining)
+    train_kwargs = {'iteration': iteration, 'pretraining': pretraining}
+    loss, predictions = model(batch, **train_kwargs)
```
gcampax: It would be more readable without the ** and the dictionary.

Pipfile
Mehrad0711: Dataparallel is buggy in torch 1.5 (https://github.com/huggingface/transformers/issues/3936)
The only known fix so far is reverting back to 1.4

gcampax (approved):
:+1:

PR #29, opened 2020-07-20
[1] Support for Translation task 

Mehrad0711:
Main changes:

- Added support for translation task and related models: Marian, MBART, T5
- Some general cleanups on paraphrasing code
- A new sentence generation method by masking tokens in the input sentence and using BART/ MBART to denoise the sentence. Can be used in single/ multi language settings and observed better results with MBART.


gcampax (commented):
I eyeballed this, but I'm going to let @s-jse do a full review because this is mostly code that touches the paraphrase part of genienlp.

Mehrad0711:
Sounds good. 
The 3 Signature mismatch warnings are irrelevant as the code in 'transformer_utils.py' is adapted from huggingface.

s-jse:
In terms of functionality, things look good. Please just

- [x]  Add a test for masked paraphrasing
- [x] Move helper functions that are specific to the translation task out of `run_generation.py`

Mehrad0711:
ok, done and done. Any other comments? Otherwise, it's good to go.

s-jse:
I think you can merge now.

PR #30, opened 2020-07-23
Wip/generalize generation

s-jse:
This PR replaces some of the generation and load code for the MQAN model family with `generate()` and `from_pretrained()` methods of the `transformers` package. This makes incorporating BART easier.


genienlp/models/general_seq2seq.py
```diff
@@ -126,7 +145,51 @@ def get_encoder_loss(self, context_rnn_state):
...
+                                     temperature=temperature,
+                                     eos_token_id=batch.decoder_vocab.eos_idx,
+                                     top_k=top_k,
+                                     top_p=top_p,
+                                     num_beams=num_beams,
+                                     no_repeat_ngram_size=no_repeat_ngram_size,
+                                     do_sample=do_sample,
+                                     generation_dict={'max_output_length': max_output_length, 'num_beams': num_beams}
+                                    )
+        generated = generated[:, 1:].cpu().apply_(self.decoder.map_to_full).to(batch.context.value.device) # remove bos token and map to full vocabulary
```
gcampax: Why do you remove the first token? The first token of the output is _not_ the init token, and you normally wouldn't remove it...

s-jse: The new generation code (from the `transformers` package) outputs BOS token as well.

Mehrad0711: I think it's better that special tokens will be handled inside numericalizer.reverse(.) which is the case already, and return full "generated" here.

Mehrad0711:
Please remove the commented lines remaining from debugging. 
Other than that looks good.

Mehrad0711 (approved):

gcampax (approved):
I think this looks sensible. Perhaps there will be some code cleanup to do later: there are some ugly methods with enormous signatures...

PR #31, opened 2020-07-23
[2] Contextual multilingual tasks 

Mehrad0711:
Changes:
- New multilingual tasks (QA, contextual NLU for user and agent) inherit from base multilingual almond task.
- Caching data is now done on-demand via --cache_input_data flag. This will free up space on k8s pods.
- Option to override context in all examples. (similar to override question; used for zero-shot cross-domain experiments)


gcampax (approved):
Yeah this looks good. Not sure what's the purpose of --override_context, but it makes sense for symmetry.

Btw, re the task naming, in the last few months I've tried to move away from the word "contextual" because the word "context" appears in too many places so it's hard to understand.

PR #34, opened 2020-08-23
Proper handling of CJK and accent characters

Mehrad0711:


gcampax (changes requested):
You cannot strip the accents, because the decoded words won't match the gold label any more when we evaluate in Genie (which does not strip the accents)

genienlp/data_utils/numericalizer/masked_tokenizer.py
```diff
@@ -124,11 +130,31 @@ def tokenize(self, tokens, mask):
                 start = end
 
             if is_bad:
-                output_tokens.append(self.unk_token)
+                # add token to extended vocab
+                if token not in self.vocab and token not in self.added_tokens_encoder:
+                    token_id = len(self.vocab) + len(self.added_tokens_encoder)
+                    self.added_tokens_encoder[token] = token_id
+                    self.added_tokens_decoder[token_id] = token
+                output_tokens.append(token)
```
gcampax, line 149: Is this adding the whole word or just the accent as a single `##` token? It would be nice not to lose the existing BERT embeddings around words with accents.

gcampax (commented):
Please do a full round of training the Thingpedia model with this before merging it.

Mehrad0711:
pending tests...

genienlp/data_utils/numericalizer/masked_tokenizer.py
Mehrad0711: addressed.

PR #37, opened 2020-09-18
Eliminate unnecessary tokenization of 'context_plus_question' field

Mehrad0711:
Let's bring back the fast data processing we deserve!
Tested on spotify_29 with 2,7784,60 train examples. Processing time reduces from 1h14m to 24m (without loss in accuracy).

### DIFF
--- a/genienlp/arguments.py
+++ b/genienlp/arguments.py
@@ -94,7 +94,7 @@ def parse_argv(parser):
                         help='whether to allow filtering on the validation sets')
     parser.add_argument('--val_batch_size', nargs='+', default=[256], type=int,
                         help='Batch size for validation corresponding to tasks in val tasks')
-
+    
     parser.add_argument('--paired', action='store_true',
                         help='Pair related examples before numericalizing the input (e.g. training with synthetic and paraphrase '
                              'sentence pairs for almond task)')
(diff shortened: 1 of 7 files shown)


genienlp/arguments.py
```diff
@@ -94,6 +94,10 @@ def parse_argv(parser):
                         help='whether to allow filtering on the validation sets')
     parser.add_argument('--val_batch_size', nargs='+', default=[256], type=int,
                         help='Batch size for validation corresponding to tasks in val tasks')
+    
+    parser.add_argument('--share_train_val_tasks', action='store_true', help='if true will override val_tasks so that'
+                                                                             ' train and val share the exact same task object')
```
gcampax: Is there any situation where we don't want this? I'd say just always do it, keep track of the task objects by name.

genienlp/data_utils/example.py
```diff
@@ -64,6 +63,9 @@ def from_raw(example_id: str, context: str, question: str, answer: str, tokenize
             args.append(words)
             args.append(mask)
         
+        args.append(args[1] + args[3])
```
gcampax, line 67: put a comment explaining this line please

gcampax (commented):
So the "unnecessary" part is only for single-turn experiments? Cause dialogues always passes 'append_question_to_context_too' so no benefit?

genienlp/arguments.py
Mehrad0711: It doesn't happen in our current experiments. But later on, if we want to try few-shot or zero-shot on tasks other than almond, we will need this option. I will change the default to be True but will keep the flag.

gcampax: No. Just make it so that tasks with the same name share the same object. There is no reason to have different objects for the same task, even if you have different lists of tasks.

Mehrad0711:
Spotify_29 is a dialog dataset. Before we were processing context_plus_question filed for each example which is wasteful since we already have processed context and question individually. This will benefit all tasks regardless of the arguments used.

gcampax:
Ok, it looks like you're still making changes to this commit so it's a bit hard for me to follow. Can you clean this up and then let me know when it's good to review?

genienlp/arguments.py
Mehrad0711: okay, that's reasonable.

Mehrad0711:
Sure, I will address your comments and ping you on slack when done.

gcampax:
Just make a single clean push to the PR and ask for review. I get email notifications...

gcampax (approved):
Yeah this looks good

PR #38, opened 2020-09-21
Wip/speedup

s-jse:
This PR makes two changes to speedup paraphrase training (only affects `genienlp train-paraphrase`) and make training GPT-large on a 16GB GPU possible.

1. If `--block_size` is provided, examples longer than that will be thrown away instead of truncated.
2. Passing `--sort_by_length` will group training examples with similar lengths together. This reduces the need for padding tokens during training and evaluation. Note that the each minibatch has examples of similar length, but the order of minibatches is still random. With batch size of 2, this has no effect on training speed; however, with larger batch sizes, the benefits are significant.
For instance, --sort_by_length gives a ~2.5x speedup for batch size of 32.
Surprisingly, this has a positive effect in terms of accuracy/BLEU score on BART,  GPT2-medium, GPT2-small and distil-GPT2, but no effect on GPT2-large.


Mehrad0711:
lgtm

Mehrad0711 (approved):

PR #39, opened 2020-09-30
Wip/fix circular import

Mehrad0711:
This fixes the circular import error caused when running paraphrase scripts.


gcampax:
What's going on here? This was supposed to be a trivial PR and it's still going? Why are we pushing reverts to a PR?

s-jse:
> 
> 
> Why are we pushing reverts to a PR?

I will squash all commits to one and merge.

PR #40, opened 2020-10-09
Some updates from SPL work

Mehrad0711:


genienlp/predict.py
```diff
@@ -54,6 +55,8 @@
 
 logger = logging.getLogger(__name__)
 
+quoted_pattern_maybe_space = re.compile(r'\"\s?([^"]*?)\s?\"')
```
gcampax: This regular expression is not correct (it will match `"foo"` without space which is not a string parameter, most importantly it will match `""` which is also not a copiable parameter.

genienlp/predict.py
```diff
@@ -184,15 +187,29 @@ def run(args, device):
...
+                    answers_wo_params.append(re.sub(quoted_pattern_maybe_space, '', a))
+                predictions_wo_params = [[] for _ in range(len(predictions))]
+                for i, preds in enumerate(predictions):
+                    for p in preds:
+                        predictions_wo_params[i].append(re.sub(quoted_pattern_maybe_space, '', p))
+
+                metrics_wo_params = calculate_and_reduce_metrics(predictions_wo_params, answers_wo_params, metrics_to_compute, args)
+                # rename em to sm
+                metrics_wo_params['sm'] = metrics_wo_params['em']
+                del metrics_wo_params['em']
```
gcampax, line 205: This is not the right metric. The right accuracy_without_params metric needs to consider all the entity tokens as well. You must use `genie evaluate-server` if you want consistent partial-accuracy metrics.

Mehrad0711: This is for getting accuracy on datasets where entities don't match anymore. e.g. directly translated eval/ test sets without alignment in SPL.
I updated the implementation by porting the requoting functions from genie.

gcampax: I guess...

gcampax (approved):

PR #46, opened 2020-11-07
BART Integration

s-jse:
This PR:
1. Changes the interface of MQAN model to make it more similar to the interface of `transformers`' `PreTrainedModel`.
1. Refactors some of validation and training methods.
1. Introduces a new task called `natural_seq2seq` for datasets that only have one input and one output (hence seq2seq) and do not require any changes to the tokenizer of the model used.
1. Adds a new model Bart to the train, prediction and server code.


genienlp/data_utils/numericalizer/transformer.py
```diff
@@ -417,3 +418,103 @@ def reverse_one(tensor):
...
+            lengths.append(len(padded_example) - max(0, 2 * max_len - len(wp_tokens_a) - len(wp_tokens_b)))
+
+            numerical.append(self._tokenizer.convert_tokens_to_ids(padded_example))
+            decoder_numerical.append([decoder_vocab.encode(word) for word in padded_example])
+
+        length = torch.tensor(lengths, dtype=torch.int32, device=device)
+        numerical = torch.tensor(numerical, dtype=torch.int64, device=device)
+        decoder_numerical = torch.tensor(decoder_numerical, dtype=torch.int64, device=device)
+
+        return SequentialField(length=length, value=numerical, limited=decoder_numerical)
```
Mehrad0711: This should be omitted right?

genienlp/data_utils/numericalizer/transformer.py
```diff
@@ -417,3 +418,103 @@ def reverse_one(tensor):
...
+            if len(tokens) == 0:
+                batch_tokens.append(' ')
+            else:
+                batch_tokens.append(' '.join(tokens))
+        
+        encoded_batch = self._tokenizer.batch_encode_plus(batch_tokens, add_special_tokens=True, pad_to_max_length=True, return_attention_masks=True)
+        length = torch.sum(torch.tensor(encoded_batch['attention_mask'], dtype=torch.int32, device=device), dim=1)
+        numerical = torch.tensor(encoded_batch['input_ids'], dtype=torch.int64, device=device)
+        # print('length = ', length)
+        # print('numerical = ', numerical)
```
Mehrad0711: Please delete commented lines intended for debugging.

genienlp/models/general_seq2seq.py
```diff
@@ -199,14 +277,98 @@ def generate(self,
...
+            y_ids = y[:, :-1].contiguous()
+            lm_labels = y[:, 1:].clone()
+            lm_labels[y[:, 1:] == pad] = -100
+            # print({'source_ids':source_ids, 'attention_mask': source_mask, 'decoder_input_ids':y_ids, 'lm_labels':lm_labels})
+            return self.bart.forward(source_ids, attention_mask=source_mask, decoder_input_ids=y_ids, lm_labels=lm_labels)
+
+            # print('input_ids = ', batch.context.value)
+            # print('lm_labels = ', batch.answer.value)
+            # return super().forward(input_ids=batch.context.value, decoder_input_ids=batch.answer.value, lm_labels=batch.answer.value)
+        else:
```
Mehrad0711: Same here.

genienlp/models/general_seq2seq.py
```diff
@@ -199,14 +277,98 @@ def generate(self,
         return generated
         
 
-    
-            
+           
+class Bart(torch.nn.Module):
```
Mehrad0711, line 281: Shouldn't Bart inherit from Seq2Seq? They share similar methods with different "forward" and perhaps "generation" function. In a sense, Seq2Seq will be our "base model" which all future models inherit from.

Mehrad0711 (changes requested):

genienlp/models/general_seq2seq.py
gcampax: Is there any advantage in inheriting from Seq2Seq? There won't be any shared code, and methods don't work the same way. I agree with Sina here, it should not inherit.

Mehrad0711: We will add new models with encoder-decoder architecture and would be nice if they share a similar interface not necessarily Seq2Seq (similar to huggingface having PreTrainedModel as the base model). We can postpone it to when we actually have more models.

gcampax: The only "interface" they need is the constructor and the "forward" method. Everything else is an implementation detail of each model. Now, I imagine Bart is actually going to be used for any Huggingface model, but that's separate.

genienlp/data_utils/numericalizer/transformer.py
s-jse: I'm changing this to a TODO. The way we handle encode_single and encode_pair should be changed to be more compatible with HuggingFace's transformers, so that we can add new models more easily.

gcampax: We should evaluate to what extent we actually need masking in the tokenizer. I suspect masking only really comes into play for @-tokens (function names / device names), and everything else can be subword tokenized freely. If that's the case, we can handle @ tokens by adding special tokens to the regular tokenizer, and let huggingface take it from there. It would be a lot faster.

Also, on the BART model we don't need to compute a separate decoder vocabulary, which means we can shave off an hour of preprocessing.

genienlp/models/general_seq2seq.py
s-jse: We will have two classes: our Seq2Seq (which is BERT-LSTM and friends) and a TrasformerSeq2Seq (which is BART, T5, etc.). We will have to wait till we update the `transformers` version to 3.5, and then we can change this Bart class to a more general TransformerSeq2Seq.

genienlp/data_utils/numericalizer/transformer.py
s-jse: Yes, I'm reevaluating mask tokens (for my other efficiency-focused PR). We can just add ThingTalk tokens to BertTokenizer's vocabulary and let it handle them. It will leave the newly added words untokenized, and is much faster. Also will let us update to FastBertTokenizer (implemented in Rust) when we update to the new `transformers` version.
But this will be a separate PR.

Mehrad0711:
Assuming Travis tests are passing locally, looks good to me.

PR #47, opened 2020-11-12
Move tokenization to before training starts

s-jse:
- Before this PR, we would tokenize training/validation examples when a new batch was requested. This PR moves tokenization (which is CPU-intensive) to before training (which is GPU-intensive). This way, we first tokenize all `Example`s into `NumericalizedExamples`s, then start training. This, enabled a few more things:
  - We can now sort the training examples based on the number of tokens that they will give to our large encoders. Previously, we were sorting based on the number of words. This reduces the number of pad tokens (i.e. wasted GPU work) and specifically improves the accuracy of BART model on the task of paraphrasing and semantic parsing.
  - Later, tokenization code can be optimized separately to levereage multiprocessing (without having to deal with CUDA and multiprocessing in Python), and/or use TokenizerFast version from transformers library (which are implemented in Rust with multithreading support).

- `--train_batch_tokens` and `--val_batch_size` are now both number of tokens by default, unless sentence_batching is used. Later we should change the name to `train_batch_size`. The default value is 4000, which fills up a 16GB GPU memory for the Schema2QA dataset, and should be increased to 5000 for MultiWOZ.

- In my experiments, models trained on this PR seem to converge after fewer iterations, so you can consider reducing number of training iterations.


gcampax:
Can we actually do this? Numericalization has to be per-batch because it clones the DecoderVocab object and modifies it in place to add any word that are outside the max_generative_vocab but can still be copied.

s-jse:
> 
> 
> Can we actually do this? Numericalization has to be per-batch because it clones the DecoderVocab object and modifies it in place to add any word that are outside the max_generative_vocab but can still be copied.

I numericalize the whole training set at once, and reuse the resulting DecoderVocab for every batch. Does this cause issues?

gcampax:
If you use the same DecoderVocab on every batch you will end up with a DecoderVocab that is as big as the full vocabulary, which means you have to pad the softmax at decode time to the full vocabulary size. That's quite slow, and will offset some of the gains in CPU time.
I don't know what the solution is here, that's why the code is still inefficient.

gcampax:
Maybe the answer is that we should not have two different vocabularies, and we should directly sample from the augmented BERT vocabulary. That way we don't have a potentially expensive conversion step.

s-jse:
I think we should -in the near future- do away with DecoderVocab and use the full BERT vocabulary for it. My experience with BART was that the model learns to copy what is relevant, and not output any of the other thousands of tokens that are available for it to copy.

gcampax:
Yeah what I worry is that you don't get that behavior unless you massively pretrain, like BART Is. Which might be fine because BART exists...

genienlp/util.py
```diff
@@ -445,21 +448,25 @@ def elapsed_time(log):
...
+                                  append_question_to_context_too=append_question_to_context_too,
+                                  override_question=override_question, override_context=override_context)
+    all_f = []
+    for i in trange(len(all_features.example_id), desc='Converting dataset to features'):
+        all_f.append(NumericalizedExamples(example_id=[all_features.example_id[i]],
+                            context=SequentialField(value=all_features.context.value[i][:all_features.context.length[i]], length=all_features.context.length[i], limited=all_features.context.limited[i][:all_features.context.length[i]]),
+                            question=SequentialField(value=all_features.question.value[i][:all_features.question.length[i]], length=all_features.question.length[i], limited=all_features.question.limited[i][:all_features.question.length[i]]),
+                            answer=SequentialField(value=all_features.answer.value[i][:all_features.answer.length[i]], length=all_features.answer.length[i], limited=all_features.answer.limited[i][:all_features.answer.length[i]]),
+                            decoder_vocab=all_features.decoder_vocab))
```
gcampax, line 463: Wait this code is suspicious. The first statement will do the actual tokenization, and the second statement is just creting the minibatches right? Do you create the minibatches only once at the beginning of the dataset?

s-jse: `NumericalizedExamples.from_examples(dataset ...` tokenizes the whole dataset to get DecoderVocab correctly. Then `all_f` breaks it down to individual examples (just reorders them in lists and removes padding).
Actual minibatches are created inside `NumericalizedExamples.collate_batches` which pads the examples again.

s-jse: Removing the pad and adding it later in `collate_batches` is a bit hacky, but has no effect on the speed since padding is really fast.

genienlp/tasks/almond/utils.py
```diff
@@ -33,7 +33,7 @@ def is_device(token):
     return token[0] == '@'
 
 def process_id(ex):
-    id_ = ex.example_id.rsplit('/', 1)
+    id_ = ex.example_id[0].rsplit('/', 1)
```
Mehrad0711: process_id is called both in BaseAlmondMultiLingualTask with Example and in Iterator when NumericalizedExample is passed. Former is passing a string whereas later passes a list. An if-else statement is needed here to handle it properly.

s-jse: Can you add it? I think the tests crashed with `id_ = ex.example_id.rsplit('/', 1)`

Mehrad0711: added.

Mehrad0711:
Rereading the code, I just noticed the shuffling is done differently now. In the previous code, we shuffled data before iterating over it. Here shuffling is done on the fly so there is a chance you see some examples more often?

s-jse:
Yes, there is a chance that we see different examples a different number of time per epoch. This is actually beneficial for BART in terms of accuracy (both on the parsing task and on the paraphrasing). It actually either keeps the accuracy the same, or very slightly improves the accuracy on other tasks as well, see "smart batching" in https://wandb.ai/pommedeterresautee/speed_training/reports/Train-HuggingFace-Models-Twice-As-Fast--VmlldzoxMDgzOTI

I've seen other repositories use sampling instead of iterating over all examples, some argue that it is closer to the i.i.d. assumptions of deep learning theory, but I don't know.

Mehrad0711:
If you train for a very long time both approaches would be equivalent (according to LLN). Since we have a fixed number of iterations I suggest doing shuffling before iteration to make sure we see the whole dataset at least once in one "epoch". On the other hand, if we run each experiment with multiple seeds (using shuffling on the fly) the max accuracy among them might be higher than shuffling before iteration.

s-jse:
The previous approach, or shuffling on the fly as you call it, was to do partial sorting within buckets of 100 batches worth of data. It results in more pad tokens in total, and has no benefits. The main reason I want to keep it this way is the accuracy boost that we empirically see for BART (and we see the exact same accuracy, if not sometimes slightly higher in my BERT-LSTM experiments).

We train for so many epochs that seeing every example once is not a problem. For Schema2QA experiments for instance, we train for 60,000 iterations, 50 examples on average per batch, and the whole dataset is ~100,000 examples. So ~30 epochs in total.
Also, our data synthesis is not guaranteed to produce a good weighting for different example types, we sometimes deduplicate and sometimes have duplicate examples in our training data, etc. This is honestly not an issue.

Mehrad0711 (approved):

genienlp/util.py
gcampax: We already do a pass over all examples in the Numericalizer to build the generatic vocab for the DecoderVocab, so now we have to do another one to build the other part of the DecoderVocab? This stuff is ripe for a cleanup. Not holding this PR for now, but definitely worth cleaning up.

PR #48, opened 2020-11-12
Replace tqdm with a line-by-line logger when not running in a terminal

gcampax:
When running in docker, or any environment where logging goes through
a file, tqdm makes a huge mess. Instead, log one line at each % change.


s-jse (approved):
Yeah, and is really useful.

PR #54, opened 2020-11-13
Wip/finetuning

Mehrad0711:
This PR introduces the following changes:

- Add fine-tuning support for Marian and MBart models.
- Includes more noising functions used in Bart and Mbart pretraining for masked paraphrasing task.
- Temporary fix in tokenization for off-by-one error in Mbart generation. There is no unique solution to this yet (https://github.com/huggingface/transformers/issues/5755)
- Introduces a method to finetune on large datasets (potentially as big as disk memory) by splitting the input file into chunks, loading them to memory one-at-a-time, and training sequentially.

### DIFF
--- a/genienlp/paraphrase/run_lm_finetuning.py
+++ b/genienlp/paraphrase/run_lm_finetuning.py
@@ -462,7 +462,7 @@ def parse_argv(parser):
     parser.add_argument("--per_gpu_train_batch_size", default=4, type=int,
                         help="Batch size per GPU/CPU for training.")
     parser.add_argument("--per_gpu_eval_batch_size", default=4, type=int,
-                        help="ze per GPU/CPU for evaluation.")
+                        help="Batch size per GPU/CPU for evaluation.")
     parser.add_argument('--gradient_accumulation_steps', type=int, default=1,
                         help="Number of updates steps to accumulate before performing a backward/update pass.")
     parser.add_argument("--learning_rate", default=5e-5, type=float,
--- a/genienlp/paraphrase/run_lm_finetuning.py
+++ b/genienlp/paraphrase/run_lm_finetuning.py
@@ -16,7 +16,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """
-This script in used for fine-tuning library models on a dataset.
+This script is used for fine-tuning library models on a dataset.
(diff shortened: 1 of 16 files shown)


gcampax (commented):
This looks okay to me, but I would want @s-jse to take a look as well. And I'd love to get rid of two copies of the training code, one for "train" and one for "run-lm-finetuning". It's a mess currently.

Mehrad0711:
Same here. The next PRs (transformer bumps) cleanup part of this code. Also now that BART is working in the main train loop, we can start merging the two copies.

genienlp/paraphrase/run_lm_finetuning.py
```diff
@@ -424,7 +462,7 @@ def parse_argv(parser):
     parser.add_argument("--per_gpu_train_batch_size", default=4, type=int,
                         help="Batch size per GPU/CPU for training.")
     parser.add_argument("--per_gpu_eval_batch_size", default=4, type=int,
-                        help="Batch size per GPU/CPU for evaluation.")
+                        help="ze per GPU/CPU for evaluation.")
```
s-jse: typo?

genienlp/paraphrase/run_lm_finetuning.py
```diff
@@ -435,7 +473,7 @@ def parse_argv(parser):
                         help="Epsilon for Adam optimizer.")
     parser.add_argument("--max_grad_norm", default=1.0, type=float,
                         help="Max gradient norm.")
-    parser.add_argument("--num_train_epochs", default=1.0, type=float,
+    parser.add_argument("--num_train_epochs", default=1, type=int,
```
s-jse, line 476: This eliminates the option to train for e.g. 1.5 epochs (I'm not sure if it worked properly before or not), but it is fine since we never used it.

s-jse (commented):
Looks good to me too. Just a typo I think.
Yes, we can and should merge training codes. I think we will have to keep generation codes (`predict.py` and `run_generation.py`) separate. We should merge paraphrasing code of run_generation.py into predict.py (it is essentially done) and let `run_generation.py` fully evolve into whatever it has become.

genienlp/paraphrase/run_lm_finetuning.py
Mehrad0711: nice catch.

Mehrad0711: Sad to bother Travis for one typo.

PR #58, opened 2020-12-01
Submission work

s-jse:
This PR:

1. Properly moves padding from tokenization to batch collation (`NumericalizedExamples.collate_batches`). Padding at collation time reduces the size of the dataset in memory (RAM).
1. Adds batching support to the server mode in `server.py`. This substantially speeds up our evaluations.
1. Fixes a previously undetected bug: when training on a dataset, the encoder embeddings should be resized so that the new words that the model can learn embeddings for the tokens that are newly added to the vocabulary.
1. During prediction, sorts examples based on their input size, and restores the original order at the end. This reduces the number of pad tokens. This especially increases the generation speed for the paraphrasing task (where if we sort by input length, we are also almost sorting by the output length, so the number of pad tokens generated in the output is also reduced).
1. `predict.py` now accepts an option `--half_precision` to use half precision for prediction. This gives a ~25% speedup on NVIDIA T4 GPUs (and V100 GPUs, both of which have specially optimized paths for 16 bit floating point operations), and reduces GPU memory consumption, so we can increase the batch size and see another speedup in addition to this.
1. Introduces dataset-specific pre- and post-processing. To see the best results using pretrained models like BART, it is necessary to make ThingTalk code as close to natural language as possible. We change the code back after generation, so this change is completely transparent.


Mehrad0711 (approved):

genienlp/arguments.py
```diff
@@ -192,6 +192,8 @@ def parse_argv(parser):
     parser.add_argument('--override_question', type=str, default=None, help='Override the question for all tasks')
     parser.add_argument('--override_context', type=str, default=None, help='Override the context for all tasks')
     parser.add_argument('--almond_preprocess_context', action='store_true', default=False, help='')
+    parser.add_argument('--almond_dataset_specific_preprocess', type=str, default='none', choices=['none', 'multiwoz'],
+                        help='Applies dataset-sepcific preprocessing to context and answer fields, and postprocesses the model outputs back to the original form.')
```
gcampax, line 196: We can merge this to simplify reproducing our submission results, but this will be on the chopping block soon.

genienlp/data_utils/example.py
```diff
@@ -78,6 +77,8 @@ class NumericalizedExamples(NamedTuple):
     question: SequentialField
     answer: SequentialField
     decoder_vocab: object
+    device: object
+    padding_function: object
```
gcampax: Do we not have a more specific type than "object"?

genienlp/data_utils/iterator.py
```diff
@@ -53,9 +53,9 @@ def __init__(self, data_source, batch_size, sort, shuffle_and_repeat, sort_key_f
         self.groups = groups
         
         if sort:
-            self.data_source = list(sorted(data_source, key=self.sort_key))
+            self.data_source, self.original_order = tuple(zip(*sorted(list(zip(data_source, range(len(data_source)))), key=lambda x: self.sort_key(x[0]))))
```
gcampax: This line is not very readable. Perhaps a comment explaining what all the zips and lists and what not do? Or even better, rewrite it in multiple clean imperative lines...

genienlp/data_utils/numericalizer/sequential_field.py
```diff
@@ -27,14 +27,14 @@
...
 import torch
 
 
 class SequentialField(NamedTuple):
-    value: torch.tensor
-    length: torch.tensor
-    limited: torch.tensor
+    value: List
+    length: List
+    limited: List
```
gcampax: List of what?

genienlp/data_utils/numericalizer/simple.py
```diff
@@ -120,33 +128,22 @@ def encode_single(self, minibatch, decoder_vocab, device=None, max_length=-1):
...
+            lengths.append(len(example))
 
-        length = torch.tensor(lengths, dtype=torch.int32, device=device)
-        numerical = torch.tensor(numerical, dtype=torch.int64, device=device)
-        decoder_numerical = torch.tensor(decoder_numerical, dtype=torch.int64, device=device)
+            numerical.append([self.vocab.stoi[word] for word in example])
+            decoder_numerical.append([decoder_vocab.encode(word) for word in example])
 
-        return SequentialField(length=length, value=numerical, limited=decoder_numerical)
+        return SequentialField(length=lengths, value=numerical, limited=decoder_numerical)
```
gcampax, line 146: Unrelated to this PR, but I'm getting the feeling that, when we kill GloVe (which we should do soon), we should also remove the whole Numericalizer abstraction and go straight to Huggingface's interface for padding. It will be more correct with models that do all sorts of weird padding things I bet.

genienlp/models/general_seq2seq.py
```diff
@@ -129,17 +129,25 @@ def __init__(self, config=None, *inputs, **kwargs):
...
+        resize_decoder: if True, will actually resize the embedding matrix of the decoder
+        """
         #logger.info(f'Vocabulary has {self.numericalizer.num_tokens} tokens from training')
         old_num_tokens = self.numericalizer.num_tokens
         new_words = []
         for task_splits in splits:
             for split in task_splits:
+                if isinstance(split, Example):
+                    # single Example
+                    split = [split]
```
gcampax: This is really ugly. Can we make the callers pass the right amount of nested lists instead? We had to fix this code like two or three times for different paths...

gcampax (commented):
Lots of small comments, nothing major

genienlp/data_utils/numericalizer/simple.py
s-jse: Yeah, we will get to that right after the `transformers` version bump, which will clean up their tokenization and padding interfaces.

genienlp/data_utils/numericalizer/transformer.py
```diff
@@ -220,35 +220,19 @@ def encode_pair(self, minibatch, decoder_vocab, device=None):
...
-                                 [self.sep_token] + \
-                                 list(wp_tokens_b[:max_len]) + \
-                                 [self.eos_token] + \
-                                 [self.pad_token] * max(0, 2 * max_len - len(wp_tokens_a) - len(wp_tokens_b))
-
-            padded.append(padded_example)
-            lengths.append(len(padded_example) - max(0, 2 * max_len - len(wp_tokens_a) - len(wp_tokens_b)))
+            example = [self.init_token] + \
+                            list(wp_tokens_a[:max_len]) + \
+                            [self.sep_token] + \
```
Mehrad0711: weirdly enough XLM-R uses two sep_tokens.

PR #59, opened 2020-12-03
Wip/transformers bump 3.5.1

Mehrad0711:
This PR bumps transformers version to 3.5.1.
Seq2Seq models now return cross-attention weights which allow substantial code refactoring and cleanup in `transformers_utils`.
Model outputs are now NamedTuples (e.g. Seq2SeqLMOutput) which allows key indexing as well as slicing.

P.S. Maybe it's time to push a new pip version!

### DIFF
--- a/genienlp/paraphrase/transformers_utils.py
+++ b/genienlp/paraphrase/transformers_utils.py
@@ -73,6 +73,9 @@ MARIAN_GROUP_MEMBERS = {
 ###############
 
 class GenieMBartTokenizer(MBartTokenizer):
+    '''
+    MBartTokenizer with the temporary fix for off-by-one error during generation: https://github.com/huggingface/transformers/issues/5755
+    '''
     vocab_files_names = {"vocab_file": "sentencepiece.bpe.model"}
     max_model_input_sizes = {m: 1024 for m in _all_mbart_models}
     pretrained_vocab_files_map = {"vocab_file": {m: SPM_URL for m in _all_mbart_models}}
@@ -99,6 +102,9 @@ class GenieMBartTokenizer(MBartTokenizer):
 
 
 class GeniePreTrainedModel(PreTrainedModel):
+    '''
+    General class for PreTrainedModel which can output cross-attention weights during generation
+    '''
     def __init__(self, config, *inputs, **kwargs):
         super().__init__(config, *inputs, **kwargs)
     
--- a/genienlp/paraphrase/transformers_utils.py
+++ b/genienlp/paraphrase/transformers_utils.py
@@ -1,10 +1,16 @@
-import copy
 import re
+import torch
+import torch.nn.functional as F
+from typing import List, Optional
 
-from transformers.modeling_bart import LayerNorm, LearnedPositionalEmbedding, BartEncoder, SelfAttention, invert_mask, \
-    SinusoidalPositionalEmbedding, BartModel, BartForConditionalGeneration
+from transformers import LogitsProcessorList
+from transformers.modeling_marian import MarianMTModel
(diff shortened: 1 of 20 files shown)


s-jse:
Since everything else (our VMs and the Pipfile here) have been upgraded to Python 3.8, should we change the Travis build configuration to also use Python 3.8, @gcampax ?

gcampax:
Ideally, we would test multiple versions of Python, because we don't want to force people who are looking to use genienlp to the very latest Python. But practically, we would run out of credits very quickly if we did that. So leaving the oldest version of Python we still support (3.6) is probably best.

gcampax:
It's also worth mentioning that we don't really use pipenv anywhere outside of Travis itself, so maybe the answer is to remove pipenv entirely...

setup.py
```diff
@@ -56,8 +56,8 @@
         'pyrouge>=0.1.3',
         'sacrebleu~=1.0',
         'requests~=2.22',
-        'transformers==2.11',
-        'sentencepiece>=0.1.83,<0.2.0',
+        'transformers==3.5.1',
+        'sentencepiece==0.1.91',
```
gcampax: Do we really need these specific versions, or can we relax the requirements a bit?

genienlp/paraphrase/transformers_utils.py
```diff
@@ -63,1096 +84,257 @@
...
-            "decoder_input_ids": decoder_input_ids,
-            "attention_mask": attention_mask,
-            "use_cache": use_cache,  # change this to avoid caching (presumably for debugging)
-        }
 
-    def _generate_no_beam_search(
+class GeniePreTrainedModel(PreTrainedModel):
+    '''
+    General class for PreTrainedModel which can output cross-attention weights during generation
+    '''
```
gcampax, line 122: I thought cross-attention weights were natively supported now?

gcampax (commented):
Generally speaking this makes sense. A couple comments. Also, I'm looking forward to more cleanups in the future.

setup.py
Mehrad0711: huggingface pins sentencepiece to 0.1.91.  Not sure if it works with other versions...

genienlp/paraphrase/transformers_utils.py
Mehrad0711: The models output cross-attention in their `forward()` method, but running `generation()` still doesn't return them. Maybe we should make a PR for this.

setup.py
gcampax: But we don't need that specific version, so you should leave our requirement more open and then pip will pick up the more stringent requirement from huggingface.

genienlp/paraphrase/transformers_utils.py
gcampax: Please do! Let's get rid of this code...

Mehrad0711: One thing though is that we need a subclass like this for returning other useful stuff such as confidence scores that `generation()` doesn't return. @s-jse may want to speak of his experience so far :)

s-jse: Yeah, we need a `GeniePreTrainedModel` class, we could get rid of the rest I think (unless a reason for model-specific behavior comes up in the future). Specifically, I need the generation code to output logits, and potentially other useful things that help us determine the model confidence.

s-jse (approved):
Looks good. I tested and the BART models we trained previously run with this version as well.

Mehrad0711:
Seems like Travis doesn't like running again. I ran the tests locally.

Issue #49, opened 2020-11-12
GPU utilization when training is low


s-jse:
GPU utilization of the BERT-LSTM model, even when we choose a batch size that fills the GPU memory is about 60% on average on a single NVIDIA V100 GPU with 16GB memory. For reference, training a BART model keeps the GPU utilization at 90% on average.

Better loading batches to GPU (maybe using multiprocessing) might help. Another potential performance bottleneck is the way we handle `DecoderVocab` objects in `MQANDecoder`.

s-jse:
Fixed by https://github.com/stanford-oval/genienlp/pull/61. It increases BERT-LSTM training speed by ~25% and GPU utilization is the same as BART, i.e. ~90% on average.

PR #61, opened 2020-12-07
The big cleanup for ThingTalk 2.0

gcampax:
This PR contains a large cleanup of genienlp, removing everything that is obsolete.

The major change is that masked tokenization is no more, because ThingTalk 2.0 doesn't use weird prefixes for parameter names or enums. Instead, the few remaining non-word tokens in ThingTalk (entities, device names) are added as special tokens to the Huggingface tokenizers, and everything else is subword tokenized. This has two practical effects:
- we greatly reduce the number of untrained embeddings
- tokenization works with any Huggingface model without code changes

GloVe and the MQAN encoder are gone. As a result, the numericalization code is cleaned up.
The internal abstraction of Seq2Seq with separate embeddings, encoder, and decoder was getting very messy, so that's gone too. There are now two models: BertLSTM and Bart.
Each model uses `--pretrained_model` to know which Huggingface model to base itself from. BertLSTM works with any Huggingface model. In principle, Bart could be T5 as well, but Huggingface doesn't have "AutoModelForConditionalGeneration" (or at least I didn't find it).

The command-line arguments are also cleaned up, which breaks backward compatibility. I think it's okay because ThingTalk broke compatibility anyway so old models don't work.
The plan is to make a 0.5.0 release before this is merged (for the sake of Schema2QA and AutoQA), then merge this, and rerun any outstanding paper and product work.

This is marked as draft because tests don't pass yet, and also because I want your opinion soon.


s-jse:
Thanks, I'll take a look at this, but have you trained anything with the new tokenization approach? How does the accuracy compare?

genienlp/arguments.py
```diff
@@ -99,9 +99,6 @@ def parse_argv(parser):
     parser.add_argument('--val_batch_size', nargs='+', default=[4000], type=int,
                         help='Number of tokens in each batch for validation, corresponding to tasks in --val_tasks')
     
-    parser.add_argument('--paired', action='store_true',
```
s-jse, line 102: We don't actively use it but I think @Mehrad0711 wanted to do experiments with it in the future.

genienlp/arguments.py
```diff
@@ -139,63 +130,20 @@ def parse_argv(parser):
     parser.add_argument("--num_beams", type=int, nargs='+', default=[1], help='1 disables beam seach')
     parser.add_argument("--no_repeat_ngram_size", type=int, nargs='+', default=[0], help='ngrams of this size cannot be repeated in the output. 0 disables it.')
 
-    parser.add_argument("--almond_has_multiple_programs", action='store_true', help='Indicate if almond dataset has multiple programs for each sentence')
+    parser.add_argument('--model', type=str, choices=['BertLSTM', 'Bart'], default='BertLSTM', help='which model to import')
+    parser.add_argument('--pretrained_model', default=None,
```
s-jse, line 128: It is not BertLSTM anymore then. TransformerLSTM maybe?

genienlp/arguments.py
```diff
@@ -99,9 +99,6 @@ def parse_argv(parser):
     parser.add_argument('--val_batch_size', nargs='+', default=[4000], type=int,
                         help='Number of tokens in each batch for validation, corresponding to tasks in --val_tasks')
     
-    parser.add_argument('--paired', action='store_true',
-                        help='Pair related examples before numericalizing the input (e.g. training with synthetic and paraphrase '
-                             'sentence pairs for almond task)')
     parser.add_argument('--max_pairs', type=int, default=1000000,
```
Mehrad0711: This should be omitted too with --paired gone.

genienlp/util.py
```diff
@@ -40,9 +40,7 @@
 import torch
 
 from .data_utils.example import NumericalizedExamples
-from .data_utils.numericalizer.sequential_field import SequentialField
 from .data_utils.iterator import LengthSortedIterator
-from .data_utils.progbar import prange
 
 logger = logging.getLogger(__name__)
```
Mehrad0711, line 46: I assume some of the amond specific pre/ post processing functions (e.g. `multiwoz_specific_preprocess` ) will be gone too with this PR?

genienlp/validate.py
```diff
@@ -47,8 +47,7 @@ def generate_with_model(model, data_iterator, numericalizer, task, args, predict
     example_ids = []
     answers = []
     contexts = []
-    questions = []
-    
+
     for batch in data_iterator:
         batch_size = len(batch.example_id)
         batch_prediction = [[] for _ in range(batch_size)] # a list where each element is a list of outputs for one input
```
Mehrad0711, line 52: if-else on line 67 should be omitted as well.

tests/test.sh
```diff
@@ -65,10 +55,10 @@ done
...
-      "--encoder_embeddings=bert-base-multilingual-uncased --decoder_embeddings= --trainable_decoder_embeddings=50 --seq2seq_encoder=Identity --dimension=768 --sentence_batching --train_batch_tokens 4 --val_batch_size 4 --use_encoder_loss --paired" \
-      "--encoder_embeddings=bert-base-multilingual-uncased --decoder_embeddings= --trainable_decoder_embeddings=50 --seq2seq_encoder=Identity --dimension=768 --rnn_zero_state cls --almond_lang_as_question" ;
+      "--model BertLSTM --pretrained_model bert-base-multilingual-uncased --trainable_decoder_embeddings=50" \
+      "--model BertLSTM --pretrained_model bert-base-multilingual-uncased --trainable_decoder_embeddings=50 --sentence_batching --train_batch_tokens 4 --val_batch_size 4 --use_encoder_loss" \
+      "--model BertLSTM --pretrained_model bert-base-multilingual-uncased --trainable_decoder_embeddings=50 --sentence_batching --train_batch_tokens 4 --val_batch_size 4 --use_encoder_loss --paired" \
```
Mehrad0711: --paired test no longer needed.

genienlp/models/base.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def init_vocab_from_data(self, vocab_sets, tasks, save_directory=None):
+        if save_directory is not None:
+            logger.info(f'Loading the accompanying numericalizer from {save_directory}')
+            self.numericalizer.load(save_directory)
+        else:
+            logger.info(f'Building vocabulary')
+            self.numericalizer.build_vocab(vocab_sets, tasks)
+
+    def add_new_vocab_from_data(self, tasks, resize_decoder=False):
```
Mehrad0711, line 68: Perhaps this method should be model specific (one for models with LSTM decoder and one for transformer seq2seq models)

genienlp/tasks/almond/__init__.py
```diff
@@ -147,58 +144,19 @@ def _detokenize_cjk_chars(self, sentence):
                 i += 1
         return "".join(output)
 
-    def tokenize(self, sentence, field_name=None):
+    def preprocess_field(self, sentence, field_name=None):
+        if self.override_context and field_name == 'context':
```
Mehrad0711, line 146: We were doing "NFD" normalization in masked_tokenizers before to reduce number of OOVs in BERT. We should add it somewhere here.  
Also `tokenize_chinese_chars` and `strip_accents` for BertTokenizer need to be True unconditionally to mimic previous behavior.

Mehrad0711 (commented):
Nice. This cleanup was much needed. A couple of comments for now:

genienlp/util.py
```diff
@@ -703,37 +633,16 @@ def load_config_json(args):
                 setattr(args, r, False)
             elif r == 'locale':
                 setattr(args, r, 'en')
-            elif r in ('trainable_decoder_embedding', 'trainable_encoder_embeddings', 'pretrain_context',
-                       'train_context_embeddings_after', 'train_question_embeddings_after'):
+            elif r == 'trainable_decoder_embedding':
                 setattr(args, r, 0)
             elif r == 'pretrain_mlm_probability':
```
s-jse: `pretrain_mlm_probability` should be removed from here too.

genienlp/util.py
```diff
@@ -703,37 +633,16 @@ def load_config_json(args):
...
-                    setattr(args, r, False)
-                else:
-                    setattr(args, r, args.train_encoder_embeddings)
-            elif r == 'train_question_embeddings':
-                setattr(args, r, args.train_encoder_embeddings)
             elif r == 'rnn_dimension':
                 setattr(args, r, args.dimension)
             elif r == 'rnn_zero_state':
                 setattr(args, r, 'zero')
             elif r == 'use_pretrained_bert':
```
s-jse, line 489: What does `use_pretrained_bert` do? Is it used anywhere?

genienlp/data_utils/numericalizer.py
```diff
@@ -0,0 +1,179 @@
...
+                self._decoder_words = [line.rstrip('\n') for line in fp]
+
+        self._init()
+
+    def pad(self, batch):
+        """
+        batch: a List of List of integers
+        """
+        #TODO account for left padding models
+        return pad_sequence(batch, padding_value=self.pad_id, batch_first=True)
```
s-jse: This unfortunately will not work with all HuggingFace models, and especially not if we want to have model-specific inputs (like BERT's `segment_id`). transformers' tokenizers don't have separate model-specific padding implementation, so this is probably the best we can do if we want to keep padding in `NumericalizedExamples.collate_batches()`.

genienlp/export.py
```diff
@@ -51,14 +52,11 @@ def main(args):
...
 
-    # we need to load the embeddings to get to the correct numericalizer class
-    # this is somewhat unfortunate but acceptable
-    numericalizer, _, _, _ = load_embeddings(args.embeddings,
-                                             args.context_embeddings,
-                                             args.question_embeddings,
-                                             args.decoder_embeddings,
-                                             args.max_generative_vocab,
-                                             logger)
+    config = AutoConfig.from_pretrained(args.pretrained_model, cache_dir=args.embeddings)
```
s-jse: Do we actually use `export.py`?

genienlp/models/bart.py
```diff
@@ -0,0 +1,99 @@
...
+                                                                 cache_dir=self.args.embeddings)
+        self.numericalizer = TransformerNumericalizer(self.args.pretrained_model, max_generative_vocab=None)
+        self.init_vocab_from_data(vocab_sets, tasks, save_directory)
+        self.bart.resize_token_embeddings(self.numericalizer.num_tokens)
+
+    def forward(self, *input, **kwargs):
+        # TODO pretraining
+        if self.training:
+            batch = input[0]
+            pretraining = kwargs.pop("pretraining", None)
```
s-jse: `pretraining` should be removed.

genienlp/models/base.py
```diff
@@ -0,0 +1,72 @@
...
+        old_num_tokens = self.numericalizer.num_tokens
+        self.numericalizer.grow_vocab(tasks)
+        if self.numericalizer.num_tokens > old_num_tokens:
+            logger.info(f'Vocabulary has expanded to {self.numericalizer.num_tokens} tokens')
+        self.bart.resize_token_embeddings(self.numericalizer.num_tokens)
```
s-jse: `self.bart`?

genienlp/models/common.py
```diff
@@ -40,11 +40,6 @@
 from torch.nn.utils.rnn import pad_packed_sequence as unpack
 
 
-class EmbeddingOutput(NamedTuple):
```
s-jse: `TransformerDecoder` and `mask_tokens()` in `common.py` seem to be unused as well.

s-jse (commented):
The changes make sense. I added a few small comments. Also, now that we are making these changes, we can also remove these last decaNLP names:
- Rename decaScore to genieScore.
- Rename `MQANDecoder` to `LSTMDecoder`.

And:
- Rename `IdentityEncoder` to `TransformerEncoder`

s-jse:
I think this might solve https://github.com/stanford-oval/genienlp/issues/49.
Also, we can now switch to HuggingFace tokenizer's `batch_encode()` since it is much faster. This, however, should be a separate PR because we should measure the tokenization speed with the new `FastTokenizer` that will be used by default in transformers version 4.0.

genienlp/arguments.py
gcampax: Yeah but the code for paired numericalization was gone before this PR.

genienlp/arguments.py
gcampax: I'd like to keep BertLSTM because that's the name we used in the papers.

genienlp/util.py
gcampax: If set to false, we load the architecture of the named bert (ie bert-base or bert-large) but then initialize it randomly. We meant to use it to see the effect of pretraining. I don't remember if we actually used it, or if it's implemented at all. Probably not.

genienlp/data_utils/numericalizer.py
gcampax: Mh there is a Tokenizer.pad method: https://huggingface.co/transformers/internal/tokenization_utils.html#transformers.tokenization_utils_base.PreTrainedTokenizerBase.pad

genienlp/export.py
gcampax: Yes we use it in the Thingpedia training code (which is currently non-functional but we need in the future)

genienlp/arguments.py
s-jse: The SPL paper calls it XLM-R-LSTM, so we could opt for a more general name.

genienlp/data_utils/numericalizer.py
s-jse: I didn't see this! It is "internal" helper, but might be a good solution.

genienlp/arguments.py
Mehrad0711: We had some improvements using --paired training in SPL experiments; but it wasn't enough to warrant its inclusion.
I prefer to have it but if it complicates the API we can drop it for now. I will do more tests and if it's conclusive we can bring it back.

genienlp/arguments.py
Mehrad0711: Yeah, I would also prefer the general term since the encoder can literally be any model from huggingface now that specific tokenizers are no longer needed.

genienlp/models/base.py
gcampax: There is nothing model specific to the interface: both models need it.

genienlp/tasks/almond/__init__.py
gcampax: Actually, we need to pass `do_basic_tokenize=False` to BertTokenizer otherwise ThingTalk symbols get all messed up, so those flags have no effect. We'll need to do things differently.

genienlp/models/base.py
Mehrad0711: Not all GenieModels have self.bart and self.decoder

gcampax: Yes I moved those bits to the subclass.

gcampax:
Ok, I merged master with Transformers 4.0 and did some testing. Unfortunately, not all fast tokenizers work correctly, because most of them don't respect whitespace and mess up identifiers and operators in ThingTalk.
Additionally, mBART needs to have the decoder replace the start token with the target language instead of using EOS, and probably needs the source language symbol as well. I imagine mT5 will need that too. I'm letting Mehrad figure all of that out.

genienlp/data_utils/example.py
```diff
@@ -27,217 +27,93 @@
...
-                     device, padding_function=numericalizer.pad)
+
+        for ex in examples:
+            yield NumericalizedExamples([ex.example_id],
+                                        numericalizer.encode_single(ex.context_plus_question),
+                                        numericalizer.encode_single(ex.answer))
 
     @staticmethod
-    def collate_batches(batches):
+    def collate_batches(batches : Iterable['NumericalizedExamples'], numericalizer, device):
```
s-jse, line 86: In `Iterable['NumericalizedExamples']` do we need `''`? Which python version introduced this?

s-jse (approved):
Looks good. `model_utils/parallel_utils.py` seems to have been unused for a while now.
On a separate note, does the code still work with multi-GPU training?

genienlp/data_utils/example.py
gcampax: We need `''` _until_ Python 3.9, because the name `NumericalizedExamples` is not in scope at that point (due to Python's weird scoping rules for classes). All versions of Python with type annotations and the typing module support this (so IIRC 3.5+)
In Python 3.9, they introduced lazy resolution of type annotations so we would not need it.

Mehrad0711:
Looks good. I need to do more tests to make sure multilingual models work and match the previous accuracy, but don't wanna hold this PR. I will send fixes if I found any problems later.

Mehrad0711 (approved):

gcampax:
I hope the code still works with multi-GPU, haven't really tested it. Anyway let's merge this and fix later I guess.

PR #66, opened 2020-12-25
Wip/version bumps

s-jse:
Version upgrades, general improvements and some nice-to-have features:
- Upgrade `transformers` to version 4.1.1.
- Upgrade `Pytorch` to version 1.7.1. Notably, this version includes this fix https://github.com/pytorch/pytorch/pull/48936 which is important for  mixed precision training.
- (bug fix) `TransformerSeq2Seq` models now use `attention_mask` properly.
- TransformerSeq2Seq models now average loss over actual sequence lengths, **excluding** pad tokens. My experiments show that this way, average log probabilities are better calibrated.
- Support "Diverse Beam Search" (https://arxiv.org/abs/1610.02424) decoding algorithm introduced in the new `transformers` version.
- Add the "Decoupled Weight Decay Regularization" (AdamW) (https://arxiv.org/abs/1711.05101) optimization algorithm.
- Add Loss Truncation (https://arxiv.org/abs/2004.14589)
- Add linear and constant learning rate schedules. Now all schedules and optimizers use `--lr_schedule` and `--lr_multiply`. No more `--transformer_lr` and `--transformer_lr_multiply`.
- Add batch encoding and tokenization using multiprocessing, and if `FastTokenizer`s are used, multithreading. This way, a 1 GB dataset is preprocessed and tokenized in only ~11 minutes on an 8-core CPU, instead of hours.
- Improved batching. Now we make sure batches never exceed the specified batch size.
- Change secondary metric of parsing tasks to F1 score instead of BLEU. BLEU is normally used for natural language tasks, not code generation.
- Improved logging messages.
- Remove some unused code.

### DIFF
--- a/genienlp/arguments.py
+++ b/genienlp/arguments.py
@@ -195,6 +195,8 @@ def post_parse(args):
         for t in args.train_task_names:
             if t not in args.val_task_names:
                 args.val_task_names.append(t)
+    if 'imdb' in args.val_task_names:
+        args.val_task_names.remove('imdb')
 
     args.timestamp = datetime.datetime.now(tz=datetime.timezone.utc).strftime('%D-%H:%M:%S %Z')
     
--- a/genienlp/arguments.py
+++ b/genienlp/arguments.py
@@ -77,21 +77,21 @@ def parse_argv(parser):
                         'If sentence_batching is used, this will be interpreted as number of examples.')
     parser.add_argument('--jump_start', default=0, type=int, help='number of iterations to give jump started tasks')
     parser.add_argument('--n_jump_start', default=0, type=int, help='how many tasks to jump start (presented in order)')
-    parser.add_argument('--num_print', default=15, type=int,
+    parser.add_argument('--num_print', default=10, type=int,
                         help='how many validation examples with greedy output to print to std out')
 
     parser.add_argument('--no_tensorboard', action='store_false', dest='tensorboard',
                         help='Turn off tensorboard logging')
     parser.add_argument('--tensorboard_dir', default=None,
                         help='Directory where to save Tensorboard logs (defaults to --save)')
-    parser.add_argument('--max_to_keep', default=5, type=int, help='number of checkpoints to keep')
-    parser.add_argument('--log_every', default=int(1e2), type=int, help='how often to log results in # of iterations')
-    parser.add_argument('--save_every', default=int(1e3), type=int,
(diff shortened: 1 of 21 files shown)


genienlp/arguments.py
```diff
@@ -192,10 +195,8 @@ def post_parse(args):
         for t in args.train_task_names:
             if t not in args.val_task_names:
                 args.val_task_names.append(t)
-    if 'imdb' in args.val_task_names:
-        args.val_task_names.remove('imdb')
 
-    args.timestamp = datetime.datetime.now(tz=datetime.timezone.utc).isoformat()
+    args.timestamp = datetime.datetime.now(tz=datetime.timezone.utc).strftime('%D-%H:%M:%S %Z')
```
gcampax, line 201: Why these two changes?
(imdb does not have validation data so the previous change was correct)

genienlp/tasks/almond/__init__.py
```diff
@@ -288,9 +292,27 @@ def _make_example(self, parts, dir_name=None, **kwargs):
...
+    def preprocess_field(self, sentence, field_name=None):
+        # remove the $ at the start of dialogue
+        # this is safe because we know we're processing dialogues, so the answer
+        # always starts with $dialogue and the context is either `null` or also
+        # starts with $dialogue
+        if field_name == 'context' and sentence.startswith('$dialogue @org.thingpedia.dialogue.transaction . '):
+            sentence = sentence[len('$dialogue @org.thingpedia.dialogue.transaction . '):]
+        if field_name == 'answer':
+            assert(sentence.startswith('$dialogue @org.thingpedia.dialogue.transaction . '))
+            sentence = sentence[len('$dialogue @org.thingpedia.dialogue.transaction . '):]
```
gcampax: Removing `$dialogue` is safe, removing `@org.thingpedia.dialogue.transaction` is not safe, because there might other dialogue policies in the future. Also, if you remove `$dialogue` entirely, you should update the comment.

genienlp/arguments.py
s-jse: I added the imdb back.
The `strftime` formatting reduces the width of log messages a bit.

genienlp/tasks/almond/__init__.py
s-jse: OK, I changed it to `$dialogue`.

s-jse:
Can we merge this now?

gcampax:
Yeah, let's merge this.

PR #67, opened 2020-12-28
Fix exporting models

gcampax:
Initialize the numericalizer correctly so we don't try to load files
that don't exist

### DIFF
--- a/genienlp/export.py
+++ b/genienlp/export.py
@@ -30,10 +30,10 @@
 import logging
 import os
 import shutil
-from transformers import AutoConfig
+import torch
 
-from .data_utils.numericalizer import TransformerNumericalizer
 from .util import load_config_json
+from . import models
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


s-jse (commented):
I guess this works, but I don't understand why we need to manually load and save the `numericalizer`.
Also, what is this export doing that `cp -r` cannot do?

gcampax:
The goal of export is to remove all the files that we don't need at inference time, such as the optimizer files and the intermediate checkpoints. There are two ways to do that: copy everything and skip files that we know we don't need, or copy only the files we know we need. Loading and saving the numericalizer copies all the numericalizer files (which we know we need), so it accomplishes the latter.

s-jse:
> 
> 
> The goal of export is to remove all the files that we don't need at inference time, such as the optimizer files and the intermediate checkpoints. There are two ways to do that: copy everything and skip files that we know we don't need, or copy only the files we know we need. Loading and saving the numericalizer copies all the numericalizer files (which we know we need), so it accomplishes the latter.

`GenieModel.from_pretrained()` already loads the numericalizer, why do we need to load it again separately?

gcampax:
Oh good catch, I meant to remove that line.

s-jse (approved):

PR #68, opened 2020-12-28
Don't load a pretrained model when loading a fully trained model

gcampax:
If we're loading an existing trained model, we don't need to also
load the pretrained model weights, because they will be overridden
immediately with the finetuned weights. This avoids an expensive
download when doing inference.


s-jse (approved):
Thanks, it was in my to-do list as well.
Can you please add a docstring to `GenieModel.from_pretrained()` and `__init__()` of the two models to explain their expected inputs?

gcampax:
Yeah I'll do that in a separate PR

PR #69, opened 2021-01-03
Wip/version bumps

s-jse:
- Adds `--dropper_min_count` argument which specifies the number of iterations after which loss truncation (if enabled) is applied.
- Fixes two issues that would occur when `--resume` is set:
  - Learning rate scheduler now actually resumes from the correct learning rate instead of restarting.
  - When loading an optimizer that was saved on a GPU machine, a GPU memory surge would happen and the run would crash. This PR solves that.


gcampax (approved):
Ok. It's a bit unfortunate that resuming has to go through all the batches again, even if it's fast.

Is the loss dropper actually useful?

s-jse:
> 
> 
> Ok. It's a bit unfortunate that resuming has to go through all the batches again, even if it's fast.
> 
> Is the loss dropper actually useful?

It is useful for factual correctness of natural language tasks like paraphrasing (and I'm hoping NLG).
It hurts semantic parsing with synthetic data, as the training loss goes to zero too quickly.
I have not tried it with noisy training data (auto-paraphrased or self-training) yet, but the claim is that it is useful for noisy training sets in general.

PR #70, opened 2021-01-03
Wip/mehrad/master

Mehrad0711:
This PR does the following:
- Adds STS-based filtering. Given a pair of text inputs (in the same or different languages) it tells you how semantically similar they are, and filters pairs with a score lower than a specified threshold.
- Adds model parallel support for training. Huggingface currently supports T5 and GPT models; BART is coming soon. It is useful for models that don't fit on one V100 gpu or fit but can't be trained with batch sizes >1 (e.g. mt5-large)
- T5 an MT5 models use sentence-piece for tokenization and its fast versions are known not to work properly.
- Bunch of updates to simplify masked paraphrasing code, fix tokenization of CJK characters, and fix a bug where crucial tokens were sometimes masked before. 
- Fix handling of CJK characters in main training code.


genienlp/data_utils/numericalizer.py
```diff
@@ -44,23 +44,23 @@
...
-    'google/mt5-xl',
-    'google/mt5-xxl',
     'sshleifer/bart-tiny-random'
 }
 # known NOT to work:
 # - all the BERT models
 # - all the XLM-R models
 #
-# mBART models work when preprocessing, because they're SPM/RoBERTa-based so they respect
+# mBART, t5, and mt5 models work when preprocessing, because they're SPM/RoBERTa-based so they respect
```
gcampax, line 53: t5 doesn't work at all if you don't preprocess, it will die if you try to add new tokens (at least in transformers 4.0)

genienlp/paraphrase/run_lm_finetuning.py
```diff
@@ -589,7 +589,8 @@ def main(args):
                                           cache_dir=args.cache_dir if args.cache_dir else None)
     tokenizer = tokenizer_class.from_pretrained(args.tokenizer_name if args.tokenizer_name else args.model_name_or_path,
                                                 do_lower_case=args.do_lower_case,
-                                                cache_dir=args.cache_dir if args.cache_dir else None)
+                                                cache_dir=args.cache_dir if args.cache_dir else None,
+                                                use_fast=False)
```
gcampax: Why use_fast=False? Without special tokens and special rules for preprocessing, fast should be good for all models, right?

genienlp/tasks/almond/__init__.py
```diff
@@ -132,6 +133,10 @@ def get_splits(self, root, **kwargs):
         return AlmondDataset.return_splits(path=os.path.join(root, 'almond'), make_example=self._make_example, **kwargs)
 
     def postprocess_answer(self, answer):
+    
+        if self._almond_detokenize_sentence:
+            answer = tokenize_cjk_chars(answer)
```
gcampax: Why is this needed? We should add a comment explaining how CJK are handled by genie and genienlp, and why.

genienlp/tasks/almond/__init__.py
```diff
@@ -166,6 +171,10 @@ def preprocess_field(self, sentence, field_name=None):
         new_sentence = ' '.join(tokens)
 
         if self._almond_detokenize_sentence:
+    
+            new_sentence = detokenize_cjk_chars(new_sentence)
+            tokens = new_sentence.split(' ')
```
gcampax: How is this code different that the previous code that detokenizes token-by-token?

genienlp/train.py
```diff
@@ -493,12 +503,41 @@ def main(args):
...
+            if any([len(val_list) == 0 for val_list in layers_list]):
+                raise ValueError('One device got no layers given the mp_device_ratio you provided'
+                                 'Hint: if you do not provide mp_device_ratio, layers will be distributed evenly')
+            
+        else:
+            n_blocks = int(math.ceil(n_layers / len(args.devices)))
+            layers_list = list(layers[i: i + n_blocks] for i in range(0, n_layers, n_blocks))
+            
+        device_map = dict(zip(args.devices, layers_list))
+        model.model.parallelize(device_map)
```
gcampax, line 541: This is TransformerSeq2Seq specific, does it work with TransformerLSTM? Or does it crash miserably?

gcampax (commented):
Overall looks good, a couple of questions

genienlp/paraphrase/run_lm_finetuning.py
Mehrad0711: Yeah this change is extreme (I was traumatized by past experiences with fast)
But I was seeing some weird tokenization issues for chinese with Marian (which is BART based).
I will add a command line flag to turn off fast.

genienlp/tasks/almond/__init__.py
Mehrad0711: It makes genienlp transparent to the changes in cjk chars spacing. 
Sure, I will add a comment.

genienlp/tasks/almond/__init__.py
Mehrad0711: The previous code wasn't adding spaces after non cjk characters. e.g. "平 巷 8 号 里" became "平巷 8号里" instead of "平巷 8 号里" (with space after 8)
I also changed run_generation to use `tokenize_cjk_chars` and `detokenize_cjk_chars` for consistency.

genienlp/train.py
Mehrad0711: No and only specific models within TransformerSeq2Seq support it so far. I will change it to raise an error in those cases.

genienlp/arguments.py
```diff
@@ -206,11 +213,22 @@ def post_parse(args):
...
-    if args.use_encoder_loss and not (args.sentence_batching and len(args.train_languages.split('+')) > 1) :
+    if args.use_encoder_loss and not (args.sentence_batching and len(args.train_languages.split('+')) > 1):
         raise ValueError('To use encoder loss you must use sentence batching and use more than one language during training.')
-
+    
+    if args.model_parallel:
+        if args.model == 'TransformerLSTM':
+            raise ValueError('Model parallel is not supported for TransformerLSTM models')
+        elif args.model == 'TransformerSeq2Seq' and args.pretrained_model not in MODEL_PARALLEL_SUPPORTED_MODELS:
+            raise ValueError('Only GPT2, T5, and MT5 models have model parallel support')
```
s-jse: Use `str(MODEL_PARALLEL_SUPPORTED_MODELS)` in the error message so that it gets updated as we add more models.

gcampax (approved):
Good for me now

genienlp/data_utils/iterator.py
```diff
@@ -100,6 +103,13 @@ def __next__(self):
                 longest_example = self.data_source[batch_of_indices[0]] # we sorted in descending order of length, so the first one is the longest
             else:
                 longest_example = new_example # this is the first element in batch, and therefore the longest
+            examples_size = self.batch_size_fn(longest_example)
+            if examples_size > self.batch_size:
+                logger.warning('Skipping an example larger than batch size. Consider increasing the batch size to avoid this warning')
+                self.last_batch_start_index += 1
+                i = self._get_next_batch_start_index()
```
s-jse: This might have to look around several times to find a short enough example.
You should just do `i += 1` and continue.

genienlp/models/transformer_seq2seq.py
```diff
@@ -44,25 +45,39 @@ def __init__(self, config=None, *inputs, args, tasks, vocab_sets, save_directory
...
         super().add_new_vocab_from_data(tasks, resize_decoder)
         self.model.resize_token_embeddings(self.numericalizer.num_tokens)
+    
+    
+    def set_decoder_start_token_id(self, lang):
+        if self._is_mbart:
+            # mBART, in contrast to MT5 or XLM-R, needs language id
+            # For now we only support single language training and evaluation with mbart models
+            lang_id = get_mbart_lang(lang)
+            self.model.config.decoder_start_token_id = self.numericalizer._tokenizer.lang_code_to_id[lang_id]
```
s-jse, line 79: Why is changing the model config necessary?

genienlp/predict.py
```diff
@@ -121,6 +121,12 @@ def run(args, device):
                                      device=device,
                                      tasks=args.tasks,
                                     )
+    
+    if args.pred_languages[0] is not None:
+        model.set_decoder_start_token_id(args.pred_languages[0].split('+')[0])
+    else:
+        # use English as default
+        model.set_decoder_start_token_id('en')
```
s-jse, line 129: Without checking if the model is mbart first? Doesn't it unintentionally affect other HuggingFace models?

genienlp/server.py
```diff
@@ -185,6 +186,8 @@ def main(args):
                                      device=device
                                     )
 
+    model.set_decoder_start_token_id(args.locale)
```
s-jse, line 189: Same here, what if it is not mbart?

genienlp/tasks/almond_utils.py
```diff
@@ -62,5 +62,34 @@ def process_id(ex):
         id_ = id_[1:]
     return id_
 
-
-
+def detokenize_cjk_chars(sentence):
```
s-jse, line 65: You are renaming this file here but not in imports. It crashes.

s-jse (commented):
Looks good. I added some comments.

genienlp/models/transformer_seq2seq.py
Mehrad0711: It checks within the `decoder_start_token_id` method if model is mbart.
For TransformerSeq2Seq models it's always a _pass_.

genienlp/predict.py
Mehrad0711: It checks within the `set_decoder_start_token_id` method if model is mbart.
For TransformerLSTM models it's a no-op.

genienlp/models/transformer_seq2seq.py
Mehrad0711: This is how huggingface training code accomplishes this. I think mainly because in generation forward pass, if `decoder_start_token_id` is not provided it will be determined inside `_get_decoder_start_token_id` in `generation_utils.py` from the config. 
It may have been used in other parts too. This being said, I think this code block can be removed (before our generation code) but need to test it first before removing it. 

```        
decoder_start_token_id = None
if self._is_mbart:
    decoder_start_token_id = self.model.config.decoder_start_token_id
````

File: genienlp/data_utils/iterator.py [hunk 12]
```diff
@@ -100,6 +103,12 @@ def __next__(self):
                 longest_example = self.data_source[batch_of_indices[0]] # we sorted in descending order of length, so the first one is the longest
             else:
                 longest_example = new_example # this is the first element in batch, and therefore the longest
+            examples_size = self.batch_size_fn(longest_example)
+            if examples_size > self.batch_size:
+                logger.warning('Skipping an example larger than batch size. Consider increasing the batch size to avoid this warning')
+                i += 1
```
s-jse, line 110: `self.last_batch_start_index` should be incremented as well, otherwise, during prediction (when `shuffle_and_repeat` is False) we might go over the same example more than once.
But even after this change, when `shuffle_and_repeat` is True, this implementation significantly increases the probability of selecting the first example that is just small enough to fit in a batch. This is fine for testing purposes. Your previous implementation which called `_get_next_batch_start_index()` was better in this regard (after adding the missing `i+=1`) .

PR #71, opened 2021-01-03
Minor fixes

s-jse:
Some additional minor fixes.
I also moved almond dataset and task definitions out of the `__init__` file.


genienlp/tasks/almond.py
```diff
@@ -32,14 +32,14 @@
...
-from .utils import ISO_to_LANG, is_device, is_entity, is_entity_marker, process_id, is_cjk_char
+from .base_task import BaseTask
+from .registry import register_task
+from .generic_dataset import CQA, default_batch_fn, input_then_output_len, input_tokens_fn
+from ..data_utils.example import Example
+from ..data_utils.progbar import progress_bar
+from .utils import is_device, is_entity, is_entity_marker, process_id, is_cjk_char
 
-from ..base_dataset import Split
+from .base_dataset import Split
```
gcampax, line 42: Is there a particular reason to move this file? The "utils.py" was meant to be almond-specific utils.

genienlp/tasks/almond.py
```diff
@@ -217,7 +217,7 @@ def _make_example(self, parts, dir_name=None, **kwargs):
             _id, sentence, target_code = parts[:3]
         else:
             _id, sentence, target_code = parts
-        question = 'translate from english to thingtalk'
+        question = ''
```
gcampax: If we ever care about multi-task learning, this will be a problem.

genienlp/tasks/almond.py
```diff
@@ -476,7 +478,6 @@ def get_splits(self, root, **kwargs):
 class AlmondMultiLingual(BaseAlmondMultiLingualTask):
     def __init__(self, name, args):
         super().__init__(name, args)
-        self.lang_as_question = args.almond_lang_as_question
```
gcampax: I remember this was helpful in multilingual setting when mixing English and foreign. @Mehrad0711 ?

genienlp/train.py
```diff
@@ -520,7 +519,7 @@ def main(args):
 
     if hasattr(args, 'tensorboard') and args.tensorboard:
         logger.info(f'Initializing Writer')
-        writer = SummaryWriter(log_dir=args.tensorboard_dir, purge_step=start_iteration)
+        writer = SummaryWriter(log_dir=args.tensorboard_dir, purge_step=start_iteration, flush_secs=60)
```
gcampax, line 523: Btw, did we ever figure out why start iteration doesn't work to erase previous failed runs?

gcampax (commented):
Mostly ok, a couple of questions

genienlp/train.py
s-jse: What do you mean? The current behavior when `--resume` is set is that tensorboard will continue from the last iteration (or best iteration if that is the checkpoint we are resuming from). The original run and the resumed one will look like this:
image

genienlp/tasks/almond.py
s-jse: Having a fixed question doesn't enable multitask learning either. We would need to set the question field as a function of other example fields.

genienlp/tasks/almond.py
s-jse: No reason at all. Do you want me to move the almond files back to a subfolder?

genienlp/train.py
gcampax: I mean when --resume is not set. In theory, passing purge_step=1 should remove all values with global_step < 1 in the same run (same tensorboard folder). In practice, the graph jumps.

genienlp/tasks/almond.py
gcampax: It's fixed per task. So it would be a different question with different tasks. That's how decanlp did it anyway (and T5 too, IIRC)

genienlp/tasks/almond.py
gcampax: No it's ok, no need to churn on top of churn. It's fine to make the change.

genienlp/tasks/almond.py
s-jse: OK, I'm reverting this change. `--overwrite_question ""` should work just fine.

genienlp/tasks/almond.py
s-jse: @Mehrad0711 said this didn't help. Some of the multilingual models we use have other means to detect languages (like a language-specific BOS token) or are trained to implicitly detect the language.
But I'm reverting this change anyways, to keep the multi-task learning option open.

Mehrad0711: Yeah, it's not really needed. Only mbart needs language specific id which is handled in this PR #70

genienlp/train.py
s-jse: Possibly related to this fix in tensorboardX: https://github.com/lanpa/tensorboardX/pull/563

s-jse: Yeah, I updated tensorboardX version and it fixed this.

gcampax (approved):
Yeah let's merge this

genienlp/tasks/almond.py
Mehrad0711: While at it can you please rename `genienlp/tasks/utils.py` to ` genienlp/tasks/almond_utils.py` ?

PR #72, opened 2021-01-04
Confidence Scoring

s-jse:
This PR adds support for calibration of neural networks (only `TransformerSeq2Seq` for now). Specifically, it adds the code to:
- Save features that are necessary for calibration. For instance, token-level and sequence level probabilities/logits before and after applying dropout (applying dropout as in https://arxiv.org/pdf/1506.02142.pdf).
- Train a boosted tree classifier using these features to predict whether or not an example is parsed correctly. The classifier's score (a number between 0 and 1) is used as a measure of confidence.
- Optionally, plot precision-recall and other useful curves.


genienlp/calibrate.py
```diff
@@ -0,0 +1,510 @@
+from typing import Callable, Iterable, List, Tuple, Union
```
gcampax, line 30: Missing copyright comment. Also, did you copy some of this code from somewhere? It looks "foreign" with the heavy functional style.

genienlp/models/transformer_seq2seq.py
```diff
@@ -159,3 +162,84 @@ def generate(self,
...
+        nodrop_logits = outputs.logits[:, :-1, :] # remove the last probability distribution which is for the token after EOS
+        for i in range(batch_size):
+            batch_nodrop_logits.append(nodrop_logits[i].gather(dim=1, index=truncated_predictions[i].view(-1, 1)).view(-1)[:prediction_lengths[i]])
+            probs = torch.softmax(nodrop_logits[i], dim=1)
+            batch_nodrop_probs.append(probs.gather(dim=1, index=truncated_predictions[i].view(-1, 1)).view(-1)[:prediction_lengths[i]])
+            batch_nodrop_entropies.append(-torch.sum(torch.log(probs)*probs, dim=1)[:prediction_lengths[i]])
+        
+        # activate dropout layers
+        should_revert_to_eval = True
+        if mc_dropout:
```
gcampax: This would be a bit more readable if "should_rever_to_eval" is set to True only in the if statement. Then later you don't need to check "mc_dropout"

genienlp/util.py
```diff
@@ -102,6 +104,70 @@ def backward(self, s: str, occurrence: str):
             
         return s
 
+
+class ConfidenceOutput:
+    """
+    Contains all necessary features that are useful for calculating confidence of a single generated output
+    """
```
gcampax, line 111: I find the naming of this class (and the flag "output_confidence") to be quite confusing. It doesn't contain the confidence, it contains the features to compute the confidence. So maybe "ConfidenceFeatures" ? I know it would be a bit confusing with the low-level features in confidence.py, but at least the confusion is contained in one file instead of being across many files and API boundaries.

genienlp/validate.py
```diff
@@ -81,21 +96,27 @@ def generate_with_model(model, data_iterator, numericalizer, task, args, predict
...
+        output = (predictions, )
+    else:
+        output = (loss, example_ids, predictions, answers, contexts)
+    if output_confidences:
+        output = output + (confidences, )
+    if output_confidence_scores:
+        confidence_scores = confidence_estimator.estimate(confidences)
+        output = output + (confidence_scores, )
+
+    return output
```
gcampax, line 120: I wonder if it's time to switch to a dataclass or namedtuple for this, it's a pretty big tuple already...

genienlp/server.py
```diff
@@ -168,6 +188,10 @@ def parse_argv(parser):
     parser.add_argument('--stdin', action='store_true', help='Interact on stdin/stdout instead of TCP')
     parser.add_argument('--locale', default='en', help='locale tag of the language to parse')
 
+    # for confidence estimation:
+    parser.add_argument('--calibrator_path', type=str, default=None, help='If provided, will be used to output confidence scores for each prediction.')
+    parser.add_argument("--mc_dropout", action='store_true', help='Monte Carlo dropout')
+    parser.add_argument("--mc_dropout_num", type=int, default=0, help='Number of samples to use for Monte Carlo dropout')
```
gcampax: What is the correct configuration to use for MC dropout? Can we make it so the estimated confidence model stores how many MC dropout samples to take (or none if dropout features are not used)? Because Genie currently has no way to pass per-model flags to server.py

genienlp/server.py
```diff
@@ -168,6 +188,10 @@ def parse_argv(parser):
     parser.add_argument('--stdin', action='store_true', help='Interact on stdin/stdout instead of TCP')
     parser.add_argument('--locale', default='en', help='locale tag of the language to parse')
 
+    # for confidence estimation:
+    parser.add_argument('--calibrator_path', type=str, default=None, help='If provided, will be used to output confidence scores for each prediction.')
```
gcampax: This should default to `--path` + /calibrator.pt or similar, and then you check if the file exists or not.

gcampax (commented):
Do you have a PR for genie-k8s with the updated pipelines?
Also, is there a way with this to have examples that are known to be OOD? I guess it would be by marking them with some pseudo-ThingTalk that is never part of the training set, so the score is always zero? (That would be `$failed ;` in ThingTalk 2.0)

genienlp/calibrate.py
s-jse: It's all mine! I'll add the usual copyright comment.

s-jse:
> 
> 
> Do you have a PR for genie-k8s with the updated pipelines?
> Also, is there a way with this to have examples that are known to be OOD? I guess it would be by marking them with some pseudo-ThingTalk that is never part of the training set, so the score is always zero? (That would be `$failed ;` in ThingTalk 2.0)

Yes, the PR is ready on k8s and yes, use $fail.

genienlp/models/transformer_seq2seq.py
s-jse: `should_revert_to_eval` and `mc_dropout` are different things. It is possible that the caller has already put the model into `train()` mode, so after tha call is done, it should still be in `train()` mode.

gcampax: If there is no mc_dropout, you don't call .train(), so you also don't need to call .eval() (that's why you check for both later).
I think it would be more readable as:
```
should_revert_to_eval = False
if mc_dropout and not self.training:
    self.train()
    should_rever_to_eval = True
```

so later you have:
```
if should_revert_to_eval:
    self.eval()
```

genienlp/server.py
s-jse: Fixed this.

genienlp/server.py
s-jse: Fixed this.

PR #74, opened 2021-01-12
Calibration changes

### LINKED ISSUE #86: Add support for combining multiple calibrators
So we can have a calibrator to identify likely wrong parses, and a calibrator to identify OOD commands.

s-jse:
This PR:
- Adds label smoothing, as https://arxiv.org/abs/1906.02629 demonstrates, label smoothing results in better calibrated models.
- Removes `--mc_dropout` argument. Now `--mc_dropout_num > 0` enables monte carlo dropout.
- Improves beam search. This is done by setting `early_stopping` to `False` when generation function is called. This way, generation will continue even after enough beams are generated, hoping that longer outputs can overtake shorter ones in terms of beam score.
- Adds more confidence features. These are primarily useful for detecting the position of errors as opposed to estimating the confidence of the whole sequence.
- Add support for multiple calibrators in `predict.py` and `server.py`.


gcampax:
I approve of the idea.

s-jse:
@gcampax This PR changes the format of the `score` field in server mode. So after this lands, genie-toolkit should change as well. The `score` field now returns a dictionary where keys are calibrator names (specified when running `genienlp calibrate`). We will have one calibrator for OOD detection and one for parse error detection.

gcampax:
> @gcampax This PR changes the format of the `score` field in server mode. So after this lands, genie-toolkit should change as well. The `score` field now returns a dictionary where keys are calibrator names (specified when running `genienlp calibrate`). We will have one calibrator for OOD detection and one for parse error detection.

Do you compute the thresholds?
Also note that we cannot change the public facing the API (the one returned by almond-nl.stanford.edu and nlp-staging.almond.stanford.edu), so we'll have to convert the two/three confidence scores into a single numeric score. We can do it in genie-toolkit or in genienlp, I have no preference, but most of the Genie code will then operate in terms of this single numeric score anyway because that's what the API returns.

gcampax (approved):
This looks good. Can you update the pipelines too?

Also, I'm going to tack on another commit that changes the KF server interface, given that we're changing the Genie side anyway. Might as well change Genie only once.

gcampax:
Also, what will the calibrator names be?

s-jse:
@gcampax , please make sure `server.py` still makes sense. Then merge this. I'll continue the calibration work after this is merged.

gcampax:
I think it does make sense. Let's merge this and test this all together.

PR #76, opened 2021-01-15
add kfserving inference to nlp server

jgd5:
This PR will make nlp server compatible with kfserving dataplane v1.

https://github.com/kubeflow/kfserving/blob/master/docs/README.md

### DIFF
--- a/dockerfiles/start.sh
+++ b/dockerfiles/start.sh
@@ -1,7 +1,6 @@
 #!/bin/bash
 set -ex
 if [ -n "${S3_MODEL_DIR}" ]; then
-  mkdir model
   aws s3 sync --exclude '*/dataset/*' --exclude '*/cache/*' --exclude 'iteration_*.pth' --exclude '*_optim.pth' "${S3_MODEL_DIR}" ./model
   genienlp kfserver --path ./model $*
 else
--- a/dockerfiles/start.sh
+++ b/dockerfiles/start.sh
@@ -1,8 +1,0 @@
-#!/bin/bash
-set -ex
-if [ -n "${S3_MODEL_DIR}" ]; then
-  aws s3 sync --exclude '*/dataset/*' --exclude '*/cache/*' --exclude 'iteration_*.pth' --exclude '*_optim.pth' "${S3_MODEL_DIR}" ./model
-  genienlp kfserver --path ./model $*
-else
-  genienlp $*
-fi
--- a/dockerfiles/start.sh
+++ b/dockerfiles/start.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+set -ex
+if [ -n "${S3_MODEL_DIR}" ]; then
+  mkdir model
+  aws s3 sync --exclude '*/dataset/*' --exclude '*/cache/*' --exclude 'iteration_*.pth' --exclude '*_optim.pth' "${S3_MODEL_DIR}" ./model
+  genienlp kfserver --path ./model $*
+else
+  genienlp $*
+fi
--- a/genienlp/kfserver.py
+++ b/genienlp/kfserver.py
@@ -1,0 +1,64 @@
+#
+# Copyright (c) 2018, Salesforce, Inc.
+#                     The Board of Trustees of the Leland Stanford Junior University
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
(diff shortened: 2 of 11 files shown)


s-jse (approved):

dockerfiles/start.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -ex
+if [ -n "${S3_MODEL_DIR}" ]; then
+  aws s3 sync --exclude '*/dataset/*' --exclude '*/cache/*' --exclude 'iteration_*.pth' --exclude '*_optim.pth' "${S3_MODEL_DIR}" ./model
+  genienlp kfserver --path ./model $*
+else
+  genienlp $*
+fi
```
gcampax: I wonder if we can avoid the start script with an init container instead. As this stands, depending on the environment variable it's a bit opaque to know if the docker image behaves normally or it behaves like the kfserver.

Also, you'll want to use `"$@"` instead of `$*`

genienlp/kfserver.py
```diff
@@ -0,0 +1,64 @@
+#
+# Copyright (c) 2018, Salesforce, Inc.
+#                     The Board of Trustees of the Leland Stanford Junior University
```
gcampax, line 3: It's a new file...

genienlp/kfserver.py
```diff
@@ -0,0 +1,64 @@
...
+    def predict(self, request):
+        results = self.server.handle_request(request)
+        return {"predictions": results}
+
+
+def main(args):
+    model, device, confidence_estimator = init(args)
+    model_server = KFModelServer('nlp', args, model.numericalizer, model, device, confidence_estimator)
+    model_server.load()
+    num_workers =  min(1, multiprocessing.cpu_count() - 1)
```
gcampax: Does this work if we're in a container? I think cpu_count looks at the actual number of physical CPUs (from /proc/cpuinfo or sysconf) but we want the number of CPUs allocated to the container. I think we need a command-line argument - or hardcode this to 1. If we do GPU inference, we'll probably need this to be 1 because GPUs cannot be shared, you hit memory issues.

gcampax (approved):
This looks good, with a few comments. Do you think there is a way to add a test for this code? Something quick, load a model, fire one or two requests at it.

dockerfiles/start.sh
jgd5: I'll look into if it's easy to add init container to the inference service spec.  If not, we can also break out kfserver into a separate Dockerfile.

genienlp/kfserver.py
jgd5: I'll hardcode to 1 for now until we have some loadtest data.

jgd5:
> This looks good, with a few comments. Do you think there is a way to add a test for this code? Something quick, load a model, fire one or two requests at it.

Is there a small model I can check into repo for testing?

jgd5:
> This looks good, with a few comments. Do you think there is a way to add a test for this code? Something quick, load a model, fire one or two requests at it.

How do I add `kfserving`  package dependency  to Travis pyenv?  I didn't see any pip install requirements.txt .  I updated setup.py but didn't seem to help.

dockerfiles/start.sh
jgd5: Looks like the spec only allows one container: https://github.com/kubeflow/kfserving/blob/master/python/kfserving/docs/V1alpha2CustomSpec.md

gcampax:
> > This looks good, with a few comments. Do you think there is a way to add a test for this code? Something quick, load a model, fire one or two requests at it.
> 
> Is there a small model I can check into repo for testing?

There is no "small" model for testing, but you can set the tests to download a good model from https://almond-static.stanford.edu/test-data/models/genienlp-v0.6.0a2.tar.xz or you can use one of the models trained during the tests.

Check how the existing tests work in test.sh

gcampax:
> > This looks good, with a few comments. Do you think there is a way to add a test for this code? Something quick, load a model, fire one or two requests at it.
> 
> How do I add `kfserving` package dependency to Travis pyenv? I didn't see any pip install requirements.txt . I updated setup.py but didn't seem to help.

Travis uses pipenv, you should edit `Pipfile` and run `pipenv lock`

jgd5:
Added a test. Travis is also happy now.

gcampax (approved):
Yeah looks great, let's merge this!

PR #79, opened 2021-01-22
Paraphrase Task

s-jse:
This PR:
- Adds a paraphrasing task to which does the necessary pre-processing and post-processing for paraphrasing. As a result of this, we can use the main prediction code (i.e. `predict.py`) to run paraphrasing.
- prediction can now be done on training set (`train.tsv` file) as well.
- Adds a progress bar to generation. It is useful for knowing the ETA of paraphrasing.

### DIFF
--- a/genienlp/tasks/almond.py
+++ b/genienlp/tasks/almond.py
@@ -270,7 +270,7 @@ class NaturalSeq2Seq(BaseAlmondTask):
 
 
 @register_task('paraphrase')
-class Paraphrase(BaseAlmondTask):
+class Paraphrase(NaturalSeq2Seq):
     """The Almond paraphrasing task. Applies the necessary preprocessing for special tokens and case changes.
     Should only be used at prediction time.
     """
@@ -282,9 +282,6 @@ class Paraphrase(BaseAlmondTask):
     @property
     def metrics(self):
         return ['bleu']
-
-    def _is_program_field(self, field_name):
-        return False
 
     def postprocess_prediction(self, example_id, prediction):
         return output_heuristics(prediction, self.reverse_maps[example_id])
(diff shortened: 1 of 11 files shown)


genienlp/tasks/almond.py
```diff
@@ -253,15 +254,63 @@ def _make_example(self, parts, dir_name=None, **kwargs):
...
-        return Example.from_raw(self.name + '/' + _id, context, question, answer,
+        return Example.from_raw(self.name + '/' + example_id, context, question, answer,
+                                preprocess=self.preprocess_field, lower=False)
+
+    def get_splits(self, root, **kwargs):
+        return AlmondDataset.return_splits(path=os.path.join(root, 'almond'), make_example=self._make_example, **kwargs)
+
+
+@register_task('paraphrase')
+class Paraphrase(BaseAlmondTask):
```
Mehrad0711: Wouldn't Paraphrase inherit from NaturalSeq2Seq task? Cause future seq2seq tasks like translation will be coming in too and they may share similar methods such as metrics, _is_program_field, etc.

Mehrad0711 (approved):
LGTM. Just a comment.
Also, do you want to update the tests in test.sh for paraphrasing?

genienlp/validate.py
```diff
@@ -61,12 +62,19 @@ def generate_with_model(model, data_iterator, numericalizer, task,args,
     answers = []
     contexts = []
 
-    for batch in data_iterator:
+    for batch in progress_bar(data_iterator, desc='Generating'):
```
gcampax: Will this affect `genienlp server` and cause us to spit out a line for every prediction?

gcampax (approved):
Looks good, just one question

genienlp/validate.py
s-jse: Yes, in stderr. Should I disable it for server mode?

gcampax: Yes please. It will generate a lot of noise otherwise, because it starts and stops a new progress bar for every sentence/minibatch.

PR #83, opened 2021-02-01
NED

Mehrad0711:
This rather small PR (as indicated by branch name) brings named entity disambiguation into genienlp.
I'm doing some final testing to make sure everything works but also want to get your early feedback on it.


gcampax:
Yep, really small pull request. Can't really review this in detail. Is this all that is needed for bootleg?

I see *a lot* of hacks. Any reference to specific domains or hardcoded types will have to go (although we can merge this specific PR first). What is the plan for that?

Also, does this work at inference time? How? Have you tested it?

s-jse:
I took a quick look at this. Where did all this Bootleg code come from? Can we just run Bootleg somewhere else (in its own repository, or at most a script here that does that) and get entities of the dataset in text format. Then everything else is normal `genienlp`?

Mehrad0711:
> I see _a lot_ of hacks. Any reference to specific domains or hardcoded types will have to go (although we can merge this specific PR first). What is the plan for that?
- We should decide what to do with domain-specific type mappings. 1) We can remove them but that would hurt performance. 2) We can continue including more domains 3) We can provide a set of all thingtalk types for all domains and filter invalid bootleg types based on that. I prefer 3) as we get the best of both worlds.

> Also, does this work at inference time? How? Have you tested it?
- Yeah. In server mode, a bootleg annotator is loaded on gpu and extracts features for each batch of examples on the fly. I also separated bootleg from the task, so we can put annotator on a server and fire examples at it (instead of loading it for each validation).

> Can we just run Bootleg somewhere else (in its own repository, or at most a script here that does that)
- This is exactly what run_bootleg.py script does by leveraging Bootleg class. This class takes care of extracting mentions, running disambiguation, collecting features, and extracting types.

s-jse:
My complaint was that this PR makes a lot of changes to the implementation of model embeddings, which were used in your experiments, but if the simpler "entities in text form" approach matched or outperformed those embedding changes, then we should just merge the simpler approach.

gcampax:
> > I see _a lot_ of hacks. Any reference to specific domains or hardcoded types will have to go (although we can merge this specific PR first). What is the plan for that?
> 
>     * We should decide what to do with domain-specific type mappings. 1) We can remove them but that would hurt performance. 2) We can continue including more domains 3) We can provide a set of all thingtalk types for all domains and filter invalid bootleg types based on that. I prefer 3) as we get the best of both worlds.

4) We can let skill developers provide the appropriate type mapping by designing a file format that Genie can generate off of Thingpedia information.

> 
> > Also, does this work at inference time? How? Have you tested it?
> 
>     * Yeah. In server mode, a bootleg annotator is loaded on gpu and extracts features for each batch of examples on the fly. I also separated bootleg from the task, so we can put annotator on a server and fire examples at it (instead of loading it for each validation).

There is no GPU in the actual NLP servers. What is the expected design here, is bootleg running in the same container as the semantic parser or no? What about the database? Do you need the Bootleg JSONs in the server container?
Are there new arguments to `genienlp server` or new environment variables (it's hard to tell from the PR)?

> My complaint was that this PR makes a lot of changes to the implementation of model embeddings, which were used in your experiments, but if the simpler "entities in text form" approach matched or outperformed those embedding changes, then we should just merge the simpler approach.

Agreed actually. It would make reviewing easier, and avoid a lot of tricky hard to test code.

Mehrad0711:
>but if the simpler "entities in text form" approach matched or outperformed those embedding changes, then we should just merge the simpler approach.

That's true for TransformerSeq2Seq models but for TransformerLSTM the text-based and non text-based approach performs similarly (on the spotify dataset I've experimented with). Besides we will need to report ablation results using different methods so I prefer if the working code is there. I have removed the non text-based code for TransformerSeq2Seq models.
But we can remove it once we start merging the code in paraphrase dir (including transformers_utils.py where the current models that use entity embedding are at) after this.

Mehrad0711:
>There is no GPU in the actual NLP servers. What is the expected design here, is bootleg running in the same container as the semantic parser or no? What about the database? Do you need the Bootleg JSONs in the server container?
Are there new arguments to genienlp server or new environment variables (it's hard to tell from the PR)?

In server mode, it runs in the same container but it can be separate. I think it should be possible to run the annotator on CPU too with lower batch size. I will test it.
And yeah bootleg model and database files are loaded from disk. (Model is ~600MB and combined database files is ~3GB)
You would need to pass `--database_dir` in genienlp (or GENIENLP_DATABASE_DIR env variable when running `genie evaluate-server`) which indicates where the files are located at.

gcampax:
> > There is no GPU in the actual NLP servers. What is the expected design here, is bootleg running in the same container as the semantic parser or no? What about the database? Do you need the Bootleg JSONs in the server container?
> > Are there new arguments to genienlp server or new environment variables (it's hard to tell from the PR)?
> 
> In server mode, it runs in the same container but it can be separate. I think it should be possible to run the annotator on CPU too with lower batch size. I will test it.

Lower batch size? Surely on CPU we can have higher batch size, not lower?

> And yeah bootleg model and database files are loaded from disk. (Model is ~600MB and combined database files is ~3GB)
> You would need to pass `--database_dir` in genienlp (or GENIENLP_DATABASE_DIR env variable when running `genie evaluate-server`) which indicates where the files are located at.

Is @jgd5 aware of these new requirements for running models?

Mehrad0711:
> Lower batch size? Surely on CPU we can have higher batch size, not lower?

Sorry I meant higher...

So if you remember, we merged some changes for serve in genie-toolkit before to run oracle models (including passing answer to genienlp server, passing --database_dir flags, etc. in predictor.ts). I have made some minor changes (renaming mainly!) which I can PR later.

gcampax:
> So if you remember, we merged some changes for serve in genie-toolkit before to run oracle models (including passing answer to genienlp server, passing --database_dir flags, etc. in preditor.ts). I have made some minor changes (renaming mainly!) which I can PR later.

Yes but Jim and I recently changed that code, so that at actual inference time (not during model evaluation, during actual deployment) genienlp is not spawned by genie-toolkit, it's run in a separate k8s container. So you need to coordinate with him if there are new arguments to add to the container.

genienlp/arguments.py
```diff
@@ -131,42 +139,90 @@ def parse_argv(parser):
...
+                        help='smaller_first: start from one token and grow into longer spans until a match is found,'
+                             'longer_first: start from the longest span and shrink until a match is found,'
+                             'ngrams: lookup all ngrams in the text and see if there is a match')
+    
+    parser.add_argument('--verbose', action='store_true', help='Print detected types for each token')
+    parser.add_argument('--almond_domains', nargs='+', default=[], help='Domains used for almond dataset; e.g. music, books, ...')
+    parser.add_argument('--features', nargs='+', type=str, default=[],
+                        help='Features that will be extracted for each entity. Order is important')
+    parser.add_argument('--features_size', nargs='+', type=int, default=[], help='Max length of each feature vector. All features are padded up to this length')
+    parser.add_argument('--features_default_val', nargs='+', type=float, default=[], help='Default value used for each feature')
```
gcampax: Please add "ner" or "ned" prefix to all the ner/ned specific arguments (at least those that don't have "bootleg" or "entity" in the name). The name is really confusing otherwise.

genienlp/data_utils/bootleg.py
```diff
@@ -0,0 +1,381 @@
...
+            hi = mid
+        else:
+            lo = mid + 1
+    return lo
+
+
+def post_process_bootleg_types(qid, type, title, almond_domains):
+    # TODO if training on multiple domains (in one run) these mapping should be modified
+    # e.g. song is mapped to book which is not correct if training on music domain too
+    for domain in almond_domains:
```
gcampax, line 66: When you train a multidomain Mario model, what domains do you pass?
This is ok for now because the PR needs to land, but you need to refactor this ASAP. Make some JSON file that Genie can generate with the information you need.

genienlp/data_utils/bootleg.py
```diff
@@ -0,0 +1,381 @@
...
+                type = 'Q2221906'
+            
+            elif 'restaurant chain' in title or title in ['restaurant', 'food manufacturer']:
+                type = 'Q571'
+            
+            elif 'writer' in title:
+                type = 'Q5'
+            
+            elif title in ['musical group', 'Wikimedia disambiguation page', 'Wikimedia list article', 'film']:
+                type = 'unk'
```
gcampax, line 93: Why is `type` a QID and not a ThingTalk type?

Why do you not traverse the wikidata graph up until you hit the right type instead of writing a bunch of if statements?

genienlp/data_utils/bootleg.py
```diff
@@ -0,0 +1,381 @@
...
+                              'association football player', 'film score composer',
+                              'disc jockey', 'record producer', 'engineer',
+                              'human biblical figure', 'big band',
+                              'musical duo', 'girl group', 'entrepreneur',
+                              'boy band', 'musical ensemble', 'artist',
+                              'vocal group', 'heavy metal band',
+                              'literary character', 'lawyer', 'lyricist',
+                              'baseball player', 'pianist', 'recording artist',
+                              'autobiographer', 'fashion designer']:
+                type = 'Q5'
```
gcampax, line 193: This is just "human". It is not right in general. How did you choose these representative QIDs?

genienlp/data_utils/bootleg.py
```diff
@@ -0,0 +1,381 @@
...
+        self.bootleg_post_process_types = args.bootleg_post_process_types
+        
+        with open(f'{self.args.database_dir}/emb_data/entityQID_to_wikidataTypeQID.json', 'r') as fin:
+            self.qid2type = ujson.load(fin)
+        with open(f'{self.args.database_dir}/es_material/type2id.json', 'r') as fin:
+            self.type2id = ujson.load(fin)
+
+        with open(f'{self.args.database_dir}/emb_data/wikidatatitle_to_typeid_0905.json', 'r') as fin:
+            title2typeid = ujson.load(fin)
+            self.typeid2title = {v: k for k, v in title2typeid.items()}
```
gcampax, line 258: What is the content of these three files? Are they just dictionaries?
If so, you can use the old [HashTable code](https://github.com/stanford-oval/genienlp/blob/69c0d5f801bf924a6a617edffcc6ca2c77971d66/genienlp/data_utils/hash_table.py) to create a binary on-disk representation of the dictionary that you can memory map. That will yield 0 memory usage, because the kernel will page in and out of disk as needed, and should solve our memory issues.

genienlp/data_utils/database_utils.py
```diff
@@ -0,0 +1,106 @@
+import unicodedata
+import re
+
+import nltk
+
+nltk.download('stopwords', quiet=True)
+from nltk.corpus import stopwords
+
+
+DOMAIN_TYPE_MAPPING = dict()
```
gcampax, line 39: Why do you need this two-level mapping instead of mapping directly to ThingTalk types in the earlier function?
In any case, this cannot be hardcoded.

genienlp/data_utils/database.py
```diff
@@ -0,0 +1,185 @@
...
+# SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import logging
+import nltk
+
+
+nltk.download('averaged_perceptron_tagger', quiet=True)
```
gcampax: Can you load this lazily, so we don't load it unless it's used?

genienlp/data_utils/bootleg.py
```diff
@@ -0,0 +1,381 @@
+import os
```
gcampax, line 31: Missing copyright notice (here and in other new files).

genienlp/data_utils/example.py
```diff
@@ -108,17 +174,25 @@ def collate_batches(batches : Iterable['NumericalizedExamples'], numericalizer,
         context_values = numericalizer.pad(context_values, pad_id=numericalizer.pad_id)
         context_limiteds = numericalizer.pad(context_limiteds, pad_id=numericalizer.decoder_pad_id)
         context_lengths = torch.stack(context_lengths, dim=0)
+        
+        if context_features:
+            # TODO change pad_id later
+            context_features = numericalizer.pad(context_features, pad_id=0)
```
gcampax: Time to fix this TODO? pad_id needs to match the real pad_id presumably?

genienlp/data_utils/iterator.py
```diff
@@ -107,12 +109,16 @@ def __next__(self):
...
             examples_size = self.batch_size_fn(longest_example)
             if examples_size > self.batch_size:
-                global _warned_for_batch_size
-                if not _warned_for_batch_size:
-                    logger.warning('Skipping an example larger than batch size. Consider increasing the batch size to avoid this warning')
-                    _warned_for_batch_size = True
+                if i not in self._skipped_examples and len(self._skipped_examples) != 0 and len(self._skipped_examples) % 10 == 0:
+                    logger.warning(f'So far, has skipped {self._skipped_examples} examples larger than batch size.'
+                                   ' Consider increasing the batch size to avoid this warning')
+                self._skipped_examples.add(i)
```
gcampax: Please restore the fact that the warning is only emitted once.

genienlp/data_utils/numericalizer.py
```diff
@@ -145,26 +167,24 @@ def save(self, save_dir):
...
             self._build_special_tokens_maps(special_tokens)
             self._build_special_tokens_regexes()
         else:
             # add the special tokens directly to the tokenizer
             self._tokenizer.add_tokens(special_tokens)
-
+        
+        # add entity boundary special tokens
+        if self.args.add_types_to_text != 'no':
+            self._tokenizer.add_tokens(['<e>', '</e>'])
```
gcampax, line 186: Have you experimented using regular tokens instead of special tokens as entity boundaries, given pretraining?

genienlp/data_utils/remote_database.py
```diff
@@ -0,0 +1,279 @@
...
+
+logger = logging.getLogger(__name__)
+
+class XPackClientMPCompatible(XPackClient):
+    def __getattr__(self, attr_name):
+        return NamespacedClient.__getattribute__(self, attr_name)
+
+class RemoteElasticDatabase(object):
+    
+    def __init__(self, config, type2id, features_default_val, features_size):
```
gcampax: How is the config passed? Is it environment variables or arguments? Given it contains secrets, it should be either a config file or environment variables.

genienlp/models/transformer_lstm.py
```diff
@@ -56,17 +58,28 @@ def __init__(self, config=None, *inputs, args, vocab_sets, tasks, save_directory
...
+        
+        self.numericalizer.get_tokenizer(save_directory)
         self.init_vocab_from_data(vocab_sets, tasks, save_directory)
 
         logger.info(f'Initializing encoder and decoder embeddings')
-        if save_directory is not None:
-            self.encoder_embeddings = AutoModel.from_config(config)
+        
+        if args.do_ner:
+            if type(config) == BertConfig:
```
gcampax, line 69: Should use `isinstance` here.

genienlp/models/transformer_lstm.py
```diff
@@ -56,17 +58,28 @@ def __init__(self, config=None, *inputs, args, vocab_sets, tasks, save_directory
...
+                self.encoder_embeddings = BertModelForNER(config, args.num_db_types, args.db_unk_id).from_pretrained(
+                    encoder_embeddings, num_db_types=args.num_db_types, db_unk_id=args.db_unk_id, cache_dir=args.embeddings)
+            elif type(config) == XLMRobertaConfig:
+                self.encoder_embeddings = XLMRobertaModelForNER(config, args.num_db_types, args.db_unk_id).from_pretrained(
+                    encoder_embeddings, num_db_types=args.num_db_types, db_unk_id=args.db_unk_id, cache_dir=args.embeddings)
```
gcampax: This code needs to handle save_directory the same way as the other branch (ie, not call from_pretrained if save_directory is set)

genienlp/server.py
```diff
@@ -45,89 +45,125 @@
...
+        all_features = NumericalizedExamples.from_examples(ex, self.numericalizer, self.args.add_types_to_text)
         # make a single batch with all examples
         return NumericalizedExamples.collate_batches(all_features, self.numericalizer, device=self.device)
+    
+    def bootleg_process_examples(self, ex, label, task):
+        line = {}
+        if task.is_contextual():
+            line['sentence'] = ex.question
+        else:
+            line['sentence'] = ex.context
```
gcampax: Wat?
If you need the task to tell you which field is the sentence, you ask the task to tell you. You don't hide this logic in a random place.
Please refactor.

genienlp/server.py
```diff
@@ -186,6 +222,7 @@ def parse_argv(parser):
                         help='Checkpoint file to use (relative to --path, defaults to best.pth)')
     parser.add_argument('--port', default=8401, type=int, help='TCP port to listen on')
     parser.add_argument('--stdin', action='store_true', help='Interact on stdin/stdout instead of TCP')
+    parser.add_argument('--database_dir', type=str, help='Database folder containing all relevant files')
```
gcampax, line 219: The database needs to be stored in the same folder as the model, because there is no way to specify a different path to the server.

genienlp/tasks/almond_task.py
```diff
@@ -1,10 +1,3 @@
-#
-# Copyright (c) 2019, The Board of Trustees of the Leland Stanford Junior University
-# All rights reserved.
-#
-# Redistribution and use in source and binary forms, with or without
-# modification, are permitted provided that the following conditions are met:
-#
```
gcampax: ???
Please don't alter copyright notices, that's literally illegal.

genienlp/tasks/almond_task.py
```diff
@@ -28,135 +21,309 @@
...
         self._almond_has_multiple_programs = args.almond_has_multiple_programs
         self._almond_detokenize_sentence = args.almond_detokenize_sentence
+        self._almond_thingtalk_version = args.almond_thingtalk_version
+        
+        self.almond_domains = args.almond_domains
+        self.all_schema_types = set()
+
+        if args.do_ner:
+            self.unk_id = args.features_default_val[0]
+            self.TTtype2DBtype = dict()
```
gcampax: As noted by the highlighting, names that start with capital letters are reserved for classes. This is not a class.

genienlp/tasks/almond_task.py
```diff
@@ -28,135 +21,309 @@
...
+            self.DBtype2TTtype = {v: k for k, v in self.TTtype2DBtype.items()}
+            self._init_db()
+
+    def _init_db(self):
+        if self.args.database_type in ['json']:
+            with open(os.path.join(self.args.database_dir, 'es_material/canonical2type.json'), 'r') as fin:
+                canonical2type = ujson.load(fin)
+            with open(os.path.join(self.args.database_dir, 'es_material/type2id.json'), 'r') as fin:
+                type2id = ujson.load(fin)
+            all_canonicals = marisa_trie.Trie(canonical2type.keys())
```
gcampax, line 84: The Marisa Trie library is designed to load/write binary files directly from disk - that's how it becomes memory efficient. Please refactor.

genienlp/tasks/almond_utils.py
```diff
@@ -78,18 +99,50 @@ def detokenize_cjk_chars(sentence):
...
+            if num_lines_in_chunk == chunk_size:
+                chunk_id += 1
+                num_lines_in_chunk = 0
+                if chunk_id == num_chunks:
+                    break
+    for file in all_out_files:
+        file.close()
+
+
+def process(args):
```
gcampax: What's this function doing? The name is too generic.

gcampax (changes requested):
There are a lot of small changes that I would like fixed before merging this, like the argument names. Then we should merge this, because the PR is already barely reviewable, and then you need to start working immediately on the major changes, because in this state I cannot use Bootleg.

genienlp/data_utils/bootleg.py
Mehrad0711: Only Spotify. Yeah, domain-specific codes will be gone (or at least reduced) in the next PR.

genienlp/data_utils/bootleg.py
Mehrad0711: That was the older implementation which was slower and had lower results for Spotify...

gcampax: Which part was the older implementation, traversing the graph? Did you investigate why?

genienlp/data_utils/bootleg.py
Mehrad0711: As long as you are training single domain (or multi-domain like Mario but you only want NED for one domain) the representative QID shouldn't matter. However, I put a TODO for multidomain (e.g. SQA multidomain) where all types should be granular so film actor --> Q10798782 , film director -->  Q2526255, etc. instead of all mapping to Q5.

gcampax: Well, the representative QID matters when you extract the Bootleg type embedding right? Otherwise you wouldn't need a QID at all.

genienlp/data_utils/bootleg.py
Mehrad0711: Yeah traversing. The dictionary that maps Qids to types is incomplete. Bootleg uses some heuristics such as comparing the frequency of a term appearing in Wikipedia to how often it appears as a labeled mention. If it appears as a mention less than <~3% of the time, they filter it out. Thus traversing doesn't always reach the type we want.

genienlp/data_utils/bootleg.py
Mehrad0711: Interesting! I'll look into it.

genienlp/data_utils/bootleg.py
gcampax: But we have the real Wikidata in Elastic Search. We can do a round of processing on the Bootleg list of types and map them to a ThingTalk type ahead of time with some ES searches. In fact, we only need to do this once and store the processed types, as long as the mapping of "root QID" to ThingTalk type doesn't change.

genienlp/data_utils/database_utils.py
Mehrad0711: Mainly for transparency. I wanted to keep everything in Q-space and only map to TT types in the final stage.
Also, in the previous embedding-based approach, bootleg_post_process_types is optional and we can just look up type Qids in type2id mapping and embed them.

genienlp/data_utils/bootleg.py
Mehrad0711: Yeah, we can use our ES to complete Bootleg databases. Will add it to the next PR's bucket list.

genienlp/data_utils/bootleg.py
Mehrad0711: Yeah, but we are not using bootleg embeddings anymore cause they didn't help. In fact, I removed all the code for integration level 2. 
(P.S. We can switch to "dump_preds" mode vs "dump_embeds" mode which saves tons of space!)

genienlp/tasks/almond_task.py
Mehrad0711: I didn't change it! It was just copied partially. Fixed now...

genienlp/data_utils/numericalizer.py
Mehrad0711: I suspect it would make a difference cause adding types to text does not generate more `correct_without_param` errors but I can try.

genienlp/data_utils/remote_database.py
Mehrad0711: It's a config file and should be located in `database_dir/es_material/elastic_config.json`

genienlp/models/transformer_lstm.py
Mehrad0711: Well I was but then realized`isinstance` checks for inheritance so `isinstance(AutoConfig.from_pretrained('xlm-roberta-base'), BertConfig)` is True!
I can reorder ifs to check for XLMRobertaConfig first, but using `type` is cleaner?

genienlp/server.py
Mehrad0711: But the database is independent of the model. Aren't we passing an ENV variable for the database?

genienlp/tasks/almond_task.py
Mehrad0711: So canonical2type is about 4G. I can create another json containing only canonical2type keys or I can save `canonical2type.keys()` to memory on the fly and load with marisa_trie. In the first case, we lose disk memory and the second approach might be slower than the current implementation? What do you think?

genienlp/server.py
Mehrad0711: Sorry, not sure what you're suggesting here. `is_contextual()` returns either True or False based on the task.
Are you suggesting it should be a class attribute rather than a method?

s-jse:
Can we finalize the genienlp version 0.6 release before merging this?

Mehrad0711:
> Can we finalize the genienlp version 0.6 release before merging this?

If you want to merge #74 before this I'm ok with it since there's not a lot of overlapping changes. Is that PR ready?

gcampax:
> Can we finalize the genienlp version 0.6 release before merging this?

As per the [release schedule](https://wiki.almond.stanford.edu/en/release-planning/two-point-oh), genienlp 0.6 will be released at the same time as Almond 2.0.
We can make more alpha releases of genienlp in meantime, if necessary. Is it necessary for any reason?

s-jse:
> 
> 
> > Can we finalize the genienlp version 0.6 release before merging this?
> 
> As per the [release schedule](https://wiki.almond.stanford.edu/en/release-planning/two-point-oh), genienlp 0.6 will be released at the same time as Almond 2.0.
> We can make more alpha releases of genienlp in meantime, if necessary. Is it necessary for any reason?

I thought a release without all the bells and whistles of bootleg would be useful for other people who may want to use genienlp for simple NLP experiments/papers.

gcampax:
Yes, we have that, it's genienlp 0.5 and lower.

genienlp/tasks/almond_task.py
Mehrad0711: Flagged for next PR.

gcampax (approved):
Yeah let's merge this and let's keep going.

PR #90, opened 2021-02-05
add kfserver builds to docker hub

jgd5:
Not sure if this will work, I was hoping docker hub builds will execute in serial or wait for the image dependency.


gcampax:
Yeah I'm not sure this will work, dockerhub build order is not reliable and it will also queue a lot of builds at once.
If we move the AWS S3 code to the `kfserver` command, the only difference left is the user we run as right? Maybe make that a build arg so we only use one Dockerfile?

jgd5:
I've updated to use one dockerfile. I also added awscli and start.sh for both.  I figure awscli is small enough and once we can use non-root user, we no longer need a kfserver specific image.

gcampax (approved):
Yeah ok

PR #93, opened 2021-02-12
NED followup (1)

Mehrad0711:
Addressing some of the TODOs in [NED project page](https://github.com/stanford-oval/genienlp/projects/2) and fixing some issues with server and numericalization.


gcampax (approved):
Ok. Does this mean that for server we run against the remote database, not the JSON database?

Mehrad0711:
Not yet but getting there. It's on the TODOs.

PR #94, opened 2021-02-12
Speed up calibration

### LINKED ISSUE #89: Calibrated models are too slow
Filing this so we can discuss what to do.
Currenlty, on CPU, a non-calibrated model replies in a few hundred milliseconds, while a calibrated model requires up to 30s to predict. Anything about 500ms is a non-starter - ideally, we'd stay under 100ms. What should we do?

@s-jse @jgd5

s-jse:
Fixes #89

After this PR, you can pass `--mc_dropout_num 0` to `genienlp predict ` and then  `--fast` to `genienlp calibrate` and get faster calibrators.

Results on a batch of 50 examples on CPU:
No calibrators:
- 6 seconds

One calibrator:
- Before this PR, with `--mc_dropout_num 5`: 44 seconds
- After removing unused confidence features, with  `--mc_dropout_num 5`: 32 seconds
- With `--mc_dropout_num 0` and `--fast` : 8 seconds

And our metric AUCPR (area under the precision-recall curve) on the almond dataset is worse by just ~0.5%.

### DIFF
--- a/genienlp/calibrate.py
+++ b/genienlp/calibrate.py
@@ -229,7 +229,8 @@ def evaluate_raw(dev_confidences: Iterable[ConfidenceFeatures], featurizer: Call
     return precision, recall, pass_rate, accuracies, thresholds
 
 def parse_argv(parser):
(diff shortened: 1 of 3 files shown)


gcampax (approved):
Sure

PR #98, opened 2021-02-19
Translation code migration

Mehrad0711:
This PR brings translation task to the main training/ prediction codebase.
A couple of things are still missing. There will be a followup PR which adds token-alignment and new translation models.  
It will also cleanup part of old code as some issues are fixed upstream on huggingface.
Have to keep transformers as git dependency for now until they do a release. This change wouldn't affect previous experiments.


genienlp/data_utils/numericalizer.py
```diff
@@ -114,27 +117,32 @@ def _use_fast(self):
...
+            tokenizer_args.update({'pretrained_model_name_or_path': save_dir, 'config': config})
         else:
-            self._tokenizer = AutoTokenizer.from_pretrained(self._pretrained_name,
-                                                            do_lower_case=False,
-                                                            do_basic_tokenize=False,
-                                                            cache_dir=self._cache,
-                                                            use_fast=self._use_fast())
-            
-        if isinstance(self._tokenizer, BertTokenizer):
+            tokenizer_args.update({'pretrained_model_name_or_path': self._pretrained_name})
```
gcampax, line 131: Do you not pass the config in this case?

genienlp/paraphrase/transformers_utils.py
```diff
@@ -106,7 +101,28 @@ def set_tgt_lang_special_tokens(self, lang: str) -> None:
...
+
+###############
+
+
+class GenieMarianTokenizer(MarianTokenizer):
+    '''
+    MarianTokenizer with the temporary fix for decoding.
+    In current huggingface's implementation `convert_tokens_to_string` method always uses spm_target to decode.
+    To be able to decode both source language and target language, this class changes the spm to be current_spm
+    '''
```
gcampax, line 114: Have we filed an issue to get this upstream?

genienlp/tasks/almond_task.py
```diff
@@ -603,11 +605,62 @@ def _make_example(self, parts, dir_name=None, **kwargs):
...
         answer = sentence # means we calculate self-bleu
         
-        return Example.from_raw(example_id, context, question, answer,
-                                preprocess=self.preprocess_field, lower=False)
+        return Example.from_raw(example_id, context, question, answer, preprocess=self.preprocess_field, lower=False)
     
-    def get_splits(self, root, **kwargs):
-        return AlmondDataset.return_splits(path=os.path.join(root, 'almond'), make_example=self._make_example, **kwargs)
+
+@register_task('translate')
```
gcampax: Should this be "almond_translate"? We already have a number of translation tasks based on WMT. Maybe we should clean that up.

gcampax (approved):
Yeah seems all straightforward. It's a bit sad that we need to so many model-specific hacks everywhere, but so be it.

genienlp/data_utils/numericalizer.py
Mehrad0711: pretrained_name is enough to load the correct tokenizer, similar to the previous code.

genienlp/paraphrase/transformers_utils.py
Mehrad0711: I will do that

genienlp/data_utils/numericalizer.py
gcampax: Then why do you need to pass the config from outside?

Mehrad0711:
I tried to minimize the amount of model-specific code and made sure most of it happens in one place (`get_tokenizer` method) so we can keep track.
Once huggingface fixes some of the issues upstream, we can remove them.
Also planning to clean up `transformers_utils.py` as generation can output attention weights in new transformers.

gcampax:
A cleanup would be nice, thank you!

genienlp/paraphrase/transformers_utils.py
Mehrad0711: Fixed upstream https://github.com/huggingface/transformers/pull/10357
Will address in the followup PR.

PR #103, opened 2021-03-03
Reduce number of train iterations in tests

s-jse:
Test time on travis is reduced from 20 minutes to 17.


Mehrad0711 (approved):
Yes, more speed!

PR #105, opened 2021-03-10
Translation code migration (2)

Mehrad0711:
With this PR, all remaining features for translation task will be in the main codebase.
If this is the case for paraphrasing task too, I think we can start removing the old code.


s-jse (approved):

gcampax (approved):
Looks ok to me

PR #106, opened 2021-03-11
batch_postprocess_prediction_ids should return target_ids

s-jse:


Mehrad0711:
I also updated tests to check for expected results to avoid future silent bugs. After only 4 train iterations, BLEU score is the only meaningful metric we can check. Since this check assumes strict reproducibility I'm not sure if the expected results need to change if, for example, we update dependencies. What do you think?
Also modified and merged similar tests for almond task since it was the bottleneck. Now tests take about ~10mins.

s-jse:
> 
> 
> I also updated tests to check for expected results to avoid future silent bugs. After only 4 train iterations, BLEU score is the only meaningful metric we can check. Since this check assumes strict reproducibility I'm not sure if the expected results need to change if, for example, we update dependencies. What do you think?
> Also modified and merged similar tests for almond task since it was the bottleneck. Now tests take about ~10mins.

The training is supposed to be deterministic, although there used to be a bug in `TransformerLSTM` models that resulted in it not being deterministic, I don't know if it is fixed. If the results are the same locally and on Travis, I would say it is safe to assume it is fixed.

If we change things like loss function (e.g. I changed the token-level averaging of loss when I added `loss_dropper`) we need to update these numbers, which is not too bad.

PR #113, opened 2021-03-15
Add write_kf_metrics command

Mehrad0711:


genienlp/write_kf_metrics.py
```diff
@@ -0,0 +1,53 @@
+# Copyright 2020 The Board of Trustees of the Leland Stanford Junior University
```
gcampax: 2021

gcampax (approved):
Yeah, reasonable. I would prefer `write-kf-metrics` with dashes, but nbd

PR #115, opened 2021-03-17
Fix NLG

s-jse:
Changes:
- Move Numericalizer's initialization code around.
- Concatenate context and question using the separator token from the Numericalizer instead of `' '`.
In the few-shot dataset, this made BART parser converge faster, but the main reason is to help the model differentiate between context and question when both are ThingTalk programs.
Parse accuracy stayed the same for both few-shot and synthetic-only training sets.


genienlp/util.py
```diff
@@ -566,11 +566,11 @@ def make_data_loader(dataset, numericalizer, batch_size, device=None, train=Fals
     else:
         return data_loader
 
-def ned_dump_entity_type_pairs(split, path, name, utterance_field):
-
-    with open(os.path.join(os.path.dirname(path), f'{name}_labels.jsonl'), 'w') as fout:
-        for ex in split:
-            text = ex.context_plus_question_with_types
+def ned_dump_entity_type_pairs(dataset, path, name, utterance_field, sep_token):
```
s-jse: Do we need to have `sep_token` here? Is it possible to instead of concatenating context and question to get `text`, dump them separately?

Mehrad0711: Yeah, we can do that. I'll push again.

gcampax (approved):
Seems a reasonable cleanup

Mehrad0711 (approved):
Tested it on dialogue datasets using NED as well, and the results are similar to what we had.

PR #124, opened 2021-04-09
Switch to bootleg 1.0.1

Mehrad0711:


genienlp/data_utils/bootleg.py
```diff
@@ -293,7 +295,68 @@ def post_process_bootleg_types(qid, type, title, almond_domains):
             elif title in ['video game', 'disease', 'city of the United States', 'taxon',
                            'Wikimedia disambiguation page', 'Wikimedia list article']:
                 type = 'unk'
+
+        elif domain == 'mario':
```
gcampax: We should avoid using the name mario anywhere outside of Kubeflow. Nobody would understand what's going on.

s-jse (approved):

PR #128, opened 2021-04-19
Add Token Classification task

Mehrad0711:
Bringing token classification task to genienlp.
Changes:
- Adding ability to generate labels per token using hf's `AutoModelForTokenClassification`
- Allow using datasets directly from hf [datasets](https://github.com/huggingface/datasets)
- Adding [cross_ner](https://paperswithcode.com/dataset/crossner) dataset
- Avoid skipping examples during validation/ prediction

PRing this before it becomes like bootleg code - big and having to survive multiple rounds of changes in genienlp!
May need some follow-up PRs after additional testing.


genienlp/models/transformer_seq2seq.py
```diff
@@ -28,6 +28,8 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 import logging
+logging.getLogger("filelock").setLevel(logging.WARNING)
```
s-jse: Isn't it enough to add it at entry points e.g. `train.py` and `predict.py`, instead of every file?

genienlp/server.py
```diff
@@ -43,7 +43,7 @@
 from .data_utils.bootleg import init_bootleg_annotator, extract_features_with_annotator
 from .tasks.registry import get_tasks
 from .util import set_seed, get_devices, load_config_json, log_model_size
-from .validate import generate_with_model
+from .validate import generate_with_seq2seq_model
```
s-jse: So token classification tasks are not supported in server mode yet?

genienlp/predict.py
```diff
@@ -218,11 +220,18 @@ def run(args, device):
             else:
                 confidence_estimators = None
             with torch.cuda.amp.autocast(enabled=args.mixed_precision):
-                generation_output = generate_with_model(model, it, model.numericalizer, task, args,
-                                                     original_order=original_order,
-                                                     output_confidence_features=args.save_confidence_features,
-                                                     confidence_estimators=confidence_estimators,
-                                                     disable_progbar=False)
+                if isinstance(model, TransformerForTokenClassification):
+                    generation_output = generate_with_classification_model(model, it, model.numericalizer, task,
```
s-jse: It would be cleaner if `generate_with_classification_model()` and `generate_with_seq2seq_model()` are combined to prevent if statements everywhere. What do you think?

genienlp/models/transformer_token_classification.py
```diff
@@ -0,0 +1,90 @@
...
+
+from ..models.base import GenieModel
+from ..util import adjust_language_code
+from . import TransformerSeq2Seq
+from ..data_utils.numericalizer import TransformerNumericalizer
+
+logger = logging.getLogger(__name__)
+
+
+class TransformerForTokenClassification(TransformerSeq2Seq, GenieModel):
```
s-jse: `TransformerSeq2Seq` is already a subclass of `GenieModel`, no need for multiple inheritance.
Also, I think more of this class's code can be inherited from its parent and reused, instead of copying their code. Can you clean it up a bit?

genienlp/models/transformer_token_classification.py
```diff
@@ -0,0 +1,90 @@
...
+            self.model = AutoModelForTokenClassification.from_pretrained(self.args.pretrained_model,
+                                                                         cache_dir=self.args.embeddings,
+                                                                         config=config)
+
+        self.numericalizer = TransformerNumericalizer(self.args.pretrained_model, args, max_generative_vocab=None,
+                                save_dir=save_directory, config=config, src_lang=self.src_lang, tgt_lang=self.tgt_lang, vocab_sets=vocab_sets, tasks=tasks)
+        
+        self.model.resize_token_embeddings(self.numericalizer.num_tokens)
+        
+        self.numericalizer.answer_pad_id = -100
```
s-jse, line 76: Is this value the same for all models? And shouldn't this code be moved to the numericalizer class?

s-jse (changes requested):
Look good overall, a few comments below. Also, remove `train_bootleg/.DS_Store` file.

genienlp/predict.py
Mehrad0711: The underlying function to get outputs are different for each model. `generate_with_seq2seq_model` uses `generate()` whereas `generate_with_classification_model` uses `forward()`. Also `cross_attentions` or `confidence` are not supported for token classification. I can move the if-else logic to a function and decide there based on model class which `generate_with` to use.

genienlp/models/transformer_token_classification.py
Mehrad0711: I need to initialize TransformerForTokenClassification from GenieModel not TransformerSeq2Seq cause it'll load unwanted things. I think `super(TransformerSeq2Seq, self).__init__(self, config)` is equivalent to `GenieModel.__init__(self, config)` but wasn't sure which one is better. Thought the second one may be more explicit. What is your suggestion?

genienlp/models/transformer_token_classification.py
Mehrad0711: `answer_pad_id` defaults to `pad_id` for all models and is set in `numericalizer.py` line 393. I overwrite it only for TransformerForTokenClassification models.

genienlp/models/transformer_seq2seq.py
Mehrad0711: Yeah In fact a better location for them is `models/__init__.py`

genienlp/models/transformer_token_classification.py
s-jse: Why is it necessary to be overridden?

genienlp/models/transformer_token_classification.py
s-jse: Why is `TransformerForTokenClassification` a subclass of `TransformerSeq2Seq` at all? In addition to the fact that it conceptually doesn't make sense (is-a relationship doesn't hold), you are not using any of the code of the parent in the child.
Isn't it better to just inherit from `GenieModel`?

genienlp/models/transformer_token_classification.py
Mehrad0711: In token classification, model generates a number in the range [0, num_labels) so it's convenient to have a different pad_id for answers other than the conventional pad_id=0

genienlp/models/transformer_token_classification.py
Mehrad0711: Hmm, yeah I think initially I was trying to reuse some methods from `TransformerSeq2Seq` like `add_new_vocab_from_data` and `confidence_features` but `add_new_vocab_from_data` is copied in and confidence is not supported yet for this model, so yeah it should be fine to use GenieModel.

genienlp/predict.py
```diff
@@ -40,6 +40,7 @@
 from torch.multiprocessing import Process, set_start_method
 
 from .data_utils.bootleg import Bootleg, init_bootleg_annotator, extract_features_with_annotator
+from .models import TransformerForTokenClassification
```
s-jse, line 43: Unused import.

s-jse (approved):

genienlp/predict.py
Mehrad0711: yeah I didn't wanna bother travis again! I'm sending a PR cleaning a bunch of things up.

Issue #91, opened 2021-02-11
Batch size should account for output length as well


s-jse:
Currently, `--train_batch_tokens ` and `--val_batch_size` both only consider the number of tokens in the input and ignore the output length. We can change this to account for both (either their sum, or their maximum, requires some benchmarking).
The reason we can make this change is that we have mainly moved to `TransformerSeq2Seq` models. Special considerations are needed to make this be efficient for `TransdormerLSTM` models as well, perhaps weighting sequence lengths by the number of parameters in encoder and decoder.

PR #130, opened 2021-04-20
Change batching

s-jse:
This PR changes the way we batch examples:
- Like before, examples are sorted based on their input (question+context) length and if tied, by output (answer) length.
However, during batching, the batch size is calculated as number of all tokens in the batch (input and output tokens, including pad tokens).
- This reduces the fluctuation in GPU memory use between different batches.
- Increase `--train_batch_tokens` and `--val_batch_size` in your experiments by a factor of 4. You should try a few values for each task to find the maximum number that does not result in OOM. Then lower the number of iterations.
For instance, `--train_batch_tokens 1500 --val_batch_size 2000` works for the `almond_dialogue_nlu` task when using BART-large (it was `--train_batch_tokens 400`). This in practice translates to a ~60% increase in the number of examples that can fit into each batch, and a ~25% decrease in speed per batch (dues to bigger batches). So overall, it improves training speed for a fixed number of examples by ~28%.
- This PR also logs the epoch number at each log step, which is (number of seen examples) / (size of the dataset).

### DIFF
--- a/genienlp/data_utils/example.py
+++ b/genienlp/data_utils/example.py
@@ -130,12 +130,15 @@ class Example(object):
     
 
 class NumericalizedExamples(NamedTuple):
+    """
+    Conatains a batch of numericalized (i.e. tokenized and converted to token ids) examples, potentially of size 1
+    """
     example_id: List[str]
     context: SequentialField
     answer: SequentialField
     
     @staticmethod
-    def from_examples(examples, numericalizer):
+    def from_examples(examples: Iterable[Example], numericalizer):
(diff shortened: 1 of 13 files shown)


s-jse:
Fixes https://github.com/stanford-oval/genienlp/issues/91

gcampax (approved):
Ok

Mehrad0711 (approved):

Mehrad0711:
Can you also update the default values for `train_batch_tokens` and `val_batch_size` in arguments.py?

PR #141, opened 2021-05-09
genienlp/server: add support for outputing multiple sequences

gcampax:
In fact, just add support for all generation hparams, because
it's easy enough to do so.

### DIFF
--- a/genienlp/server.py
+++ b/genienlp/server.py
@@ -102,18 +102,36 @@ class Server(object):
         with torch.no_grad():
             if self.args.calibrator_paths is not None:
                 output = generate_with_model(self.model, [batch], self.numericalizer, task, self.args,
-                                                        output_predictions_only=True,
-                                                        confidence_estimators=self.confidence_estimators)
+                                             output_predictions_only=True,
+                                             confidence_estimators=self.confidence_estimators)
                 response = []
-                for idx, p in enumerate(output.predictions):
-                    instance = {'answer': p[0], 'score': {}}
-                    for e_idx, estimator_scores in enumerate(output.confidence_scores):
-                        instance['score'][self.estimator_filenames[e_idx]] = float(estimator_scores[idx])
-                    response.append(instance)
+                if sum(self.args.num_outputs) > 1:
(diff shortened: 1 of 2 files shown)


genienlp/server.py
```diff
@@ -102,18 +102,36 @@ def handle_request(self, request):
...
-                                                        confidence_estimators=self.confidence_estimators)
+                                             output_predictions_only=True,
+                                             confidence_estimators=self.confidence_estimators)
                 response = []
-                for idx, p in enumerate(output.predictions):
-                    instance = {'answer': p[0], 'score': {}}
-                    for e_idx, estimator_scores in enumerate(output.confidence_scores):
-                        instance['score'][self.estimator_filenames[e_idx]] = float(estimator_scores[idx])
-                    response.append(instance)
+                if len(self.args.num_outputs) > 1 or self.args.num_outputs[0] > 1:
```
s-jse: The correct condition would be `sum(self.args.num_outputs) > 1` since it is always an array.

genienlp/server.py
```diff
@@ -102,18 +102,36 @@ def handle_request(self, request):
...
+                    for idx, predictions in enumerate(output.predictions):
+                        candidates = []
+                        for cand in predictions:
+                            candidate = {'answer': cand, 'score': {}}
+                            for e_idx, estimator_scores in enumerate(output.confidence_scores):
+                                candidate['score'][self.estimator_filenames[e_idx]] = float(estimator_scores[idx])
+                            candidates.append(candidate)
+                        response.append({'candidates': candidates})
+                else:
```
s-jse, line 118: It may make things cleaner if we always return a list. But I leave it up to you.

s-jse (changes requested):
Thanks. I added two comments, and one test is also failing.

genienlp/server.py
gcampax: I wanted to avoid accidentally introducing regressions, by keeping the existing code.

s-jse:
OK, let's merge this.

PR #142, opened 2021-05-11
Fix paraphrasing

gcampax:
It's broken in Kubeflow


s-jse (approved):
Looks good if it works on kubeflow.

gcampax:
The generate paraphrase part of the pipeline seems to work. Assuming prediction (filtering) also works, I'm going to merge this.

PR #146, opened 2021-05-24
Wip/format

Mehrad0711:
Let's bring some style to genienlp!
This PR autoformats all files using [isort](https://github.com/PyCQA/isort) and [black](https://github.com/psf/black). The pre-commit hooks will take care of style checking for all modified files before committing the code. We can tweak the configs to only keep the changes we like. (black is not very configurable though due to their philosophy of being uncompromising!)


gcampax (approved):
Looks good to me

PR #162, opened 2021-06-28
Misc. updates

Mehrad0711:
Changes:
- Adding several new metrics useful for scoring NLG results.
- write_kf_metrics can now parse multiple eval files. It also writes UI metadata used for visualizing results.
- Bootleg has the option to prep test set too which cuts down prediction time.
- Improve alignment heuristics used in translation.

### DIFF
--- a/genienlp/arguments.py
+++ b/genienlp/arguments.py
@@ -121,6 +121,15 @@ def parse_argv(parser):
     parser.add_argument(
         '--num_print', default=10, type=int, help='how many validation examples with greedy output to print to std out'
     )
+
+    parser.add_argument(
+        '--override_valid_metrics',
+        nargs='+',
+        action='append',
+        type=str,
+        help='if provided will override metrics provided by the task (format is a list of lists)',
+    )
+
     parser.add_argument(
         '--print_train_examples_too',
         action='store_true',
@@ -651,6 +660,12 @@ def post_parse_train_specific(args):
             'You should use the cased version of provided model when not preprocessing special tokens.'
             ' Otherwise the program (answer) will not be tokenized properly'
         )
+
+    if args.override_valid_metrics:
+        assert len(args.override_valid_metrics) == len(args.train_tasks) == len(args.val_tasks)
+        for train_task, val_task, metrics in zip(args.train_tasks, args.val_tasks, args.override_valid_metrics):
+            train_task.metrics = metrics
+            val_task.metrics = metrics
 
     args.log_dir = args.save
     if args.tensorboard_dir is None:
--- a/genienlp/predict.py
+++ b/genienlp/predict.py
@@ -261,6 +261,7 @@ def run(args, device):
 
             if len(generation_output.answers) > 0:
                 metrics_to_compute = task.metrics
+                metrics_to_compute += args.extra_metrics
                 if args.main_metric_only:
                     metrics_to_compute = [metrics_to_compute[0]]
                 metrics = calculate_and_reduce_metrics(
@@ -314,6 +315,7 @@ def parse_argv(parser):
         help='Name of dataset to run prediction for; will be ignored if --evaluate is test',
     )
     parser.add_argument('--tasks', dest='task_names', nargs='+', help='task names for prediction')
+    parser.add_argument('--extra_metrics', nargs='+', default=[], help='include these additional metrics in reported results')
     parser.add_argument(
         '--devices',
         default=None,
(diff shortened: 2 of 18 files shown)


genienlp/arguments.py
```diff
@@ -652,6 +661,12 @@ def post_parse_train_specific(args):
             ' Otherwise the program (answer) will not be tokenized properly'
         )
 
+    if args.override_valid_metrics:
+        assert len(args.override_valid_metrics) == len(args.train_tasks) == len(args.val_tasks)
+        for train_task, val_task, metrics in zip(args.train_tasks, args.val_tasks, args.override_valid_metrics):
+            train_task.metrics = metrics
+            val_task.metrics = metrics
```
gcampax, line 668: Does this code work at all?
`override_valid_metrics` is a list of strings, so within the loop `metrics` is a single string, but `task.metrics` should be a list of strings.

genienlp/predict.py
```diff
@@ -445,6 +445,16 @@ def parse_argv(parser):
         help='if true the provided dataset would not contain the answer (translated sentence)',
     )
     parser.add_argument('--plot_heatmaps', action='store_true', help='whether to plot cross-attention heatmaps')
+    parser.add_argument(
+        '--replace_qp',
```
gcampax, line 449: This flag is not very understandable, can we choose a better name?
Also, rather than have two flags, maybe have one flag with parameter `yes`, `no`, `force`?

genienlp/run_bootleg.py
```diff
@@ -182,6 +187,9 @@ def parse_argv(parser):
     parser.add_argument(
         '--almond_domains', nargs='+', default=[], help='Domains used for almond dataset; e.g. music, books, ...'
     )
+
+    parser.add_argument('--prep_test_too', action='store_true', help='Prepare bootleg features for test set too')
```
gcampax: `--data_splits {train|dev|test}` or similar would be cleaner...

genienlp/util.py
```diff
@@ -759,9 +755,18 @@ def load_config_json(args):
             'reduce_metrics',
             'database_dir',
         ]
+        # these are true/ false arguments
+        overwrite_actions = [
+            'plot_heatmaps',
```
gcampax: Why do you want to copy "plot_heatmaps" from training? It doesn't make sense...

tests/test_translation.sh
```diff
@@ -11,10 +11,10 @@ for model in "Helsinki-NLP/opus-mt-en-de" "sshleifer/tiny-mbart" ; do
 
     if [[ $model == Helsinki-NLP* ]] ; then
       base_model="marian"
-      expected_result='{"bleu": 95.37006954777766}'
+      expected_result='{"sacrebleu": 88.04086004116694}'
```
gcampax: Why is the metric called sacrebleu now? sacrebleu is just the implementation, the metric is still bleu.

genienlp/util.py
Mehrad0711: fixed

genienlp/arguments.py
Mehrad0711: So `override_valid_metrics` is list of list of strings (note nargs='+' and action='append'). If you have multiple tasks, you would use this flag as many times.

genienlp/predict.py
Mehrad0711: I'm actually filing a PR after this which changes the way we do translation + alignment. I'm removing the `force_replace_qp` there. As that PR is backward incompatible by nature, I would do the renaming there and update all the k8s pipelines.

genienlp/run_bootleg.py
Mehrad0711: sure

tests/test_translation.sh
Mehrad0711: For backward compatibility, I didn't wanna change the definition of "our" bleu score. The only difference between `blue` and `sacrebleu` is that `sacrebleu` does not do lowercasing so it's appropriate for NLG or other tasks that preserev casing. Otherwise, their implementation is the same. If you all prefer, I can change "bleu" to not lowercase and remove "sacrebleu". Just beware that new bleu scores on non-almond datasets will be lower.

gcampax: Maybe "casedbleu" then? "sacrebleu" is not a meaningful metric name, it doesn't tell you what's the difference with normal bleu

Mehrad0711: okay

gcampax (approved):
LGTM

Issue #156, opened 2021-06-16
Bootleg takes way too much space and memory


gcampax:
I had to bump up the ephemeral storage of BART models with Bootleg to 55G. It takes ~20 minutes for a model to start for inference, as it downloads and then loads in memory gigs and gigs of "stuff". This is simply not feasible.

Mehrad0711:
Addressed in #166

gcampax:
No it wasn't! You only addressed the memory part (and only in the sense that memory is reclaimable, not in the sense that memory is not actually used), startup time and disk usage are still to be fixed.

Mehrad0711:
I'm afraid we need to add more assignees then! Addressing disk space requires more than surface changes in bootleg code.

PR #166, opened 2021-06-29
Use wip/neural branch for bootleg

Mehrad0711:
Changes to bootleg:
- Fixing memory issues by memory mapping all large files (addressing #156)
- hotfix for missing titles for certain Qids
- fix device assertion

Have to bring git dependencies back until bootleg pushes a new release.
Code is tested on kubeflow using thingpedia 121 dataset.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -65,7 +65,7 @@ setuptools.setup(
         # for kf
         'kfserving>=0.5.0',
         # for NED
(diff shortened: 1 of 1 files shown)


gcampax (approved):
Yeah we talked about this on Slack, go ahead.

PR #168, opened 2021-06-30
Revamp translation approach

Mehrad0711:
These changes will improve the translation quality both with and without alignment. Please see the commit messages for the details.
I had to keep the old translation code so `run-paraphrase` command still work. Once we remove the old codebase, we can remove those too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ genienlp run-paraphrase --model_name_or_path <model_dir> --temperature 0.3 --rep
 Use the following command for training/ finetuning an NMT model:
 
 ```bash
-genienlp train train --train_tasks almond_translate --data <data_directory> --train_languages <src_lang> --eval_languages <tgt_lang> --no_commit --train_iterations <iterations> --preserve_case --save <save_dir> --exist_ok --skip_cache --model TransformerSeq2Seq --pretrained_model <hf_model_name>
+genienlp train --train_tasks almond_translate --data <data_directory> --train_languages <src_lang> --eval_languages <tgt_lang> --no_commit --train_iterations <iterations> --preserve_case --save <save_dir> --exist_ok --skip_cache --model TransformerSeq2Seq --pretrained_model <hf_model_name>
 ```
 
 We currently support MarianMT, MBART, MT5, and M2M100 models.<br>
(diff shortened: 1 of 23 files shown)


README.md
```diff
@@ -117,19 +117,39 @@ Generate paraphrases:
 genienlp run-paraphrase --model_name_or_path <model_dir> --temperature 0.3 --repetition_penalty 1.0 --num_samples 4 --batch_size 32 --input_file <input_tsv_file> --input_column 1
 ```
 
+### Translation
+
+Use the following command for training/ finetuning an NMT model:
+
+```bash
+genienlp train train --train_tasks almond_translate --data <data_directory> --train_languages <src_lang> --eval_languages <tgt_lang> --no_commit --train_iterations <iterations> --preserve_case --save <save_dir> --exist_ok --skip_cache --model TransformerSeq2Seq --pretrained_model <hf_model_name>
```
s-jse: typo

genienlp/data_utils/example.py
```diff
@@ -88,19 +89,23 @@ def __init__(
         self,
         example_id: str,
         context: str,
+        context_spans: List[int],
```
s-jse: Is there a way to avoid adding `*spans` everywhere in the code?

genienlp/tasks/almond_task.py
```diff
@@ -604,6 +604,22 @@ def _make_example(self, parts, dir_name=None, **kwargs):
...
 
+def split_text_into_sentences(text, lang):
+    import re
+
+    if lang in ['zh', 'ja', 'ko']:
+        sentences = list(re.findall(u'([^!?。]+[!?。]*)\s?', text, flags=re.U))
+    elif lang in ['en']:
+        sentences = re.sub('(?<!\w\.\w.)(?<![A-Z][a-z]\.)(?<=[\.\!\?])\s|(?:。)', '\n', text).split('\n')
+    else:
+        import nltk
```
s-jse, line 615: why not use `nltk` for English as well? Also, it has neural models for sentence boundary detection, does this perform as well as those neural models?

s-jse (approved):

genienlp/data_utils/example.py
Mehrad0711: This is the fast/ clean way. Otherwise, you have to do multiple tokenization encoding and decodings along the way to make sure 1) you know where the spans are in source language 2) sentences are translated without quotation 3) you know where to put them back in the translated sentences, which would be much slower.

genienlp/tasks/almond_task.py
Mehrad0711: I tried nltk's "PunktSentenceTokenizer" and unfortunately it doesn't work well for edge cases.
For example: `I was going there...  But now I'm here, i.e. present`
would be tokenized as `["I was going there...  But now I'm here, i.e.", 'present']` using nltk which is wrong for two reasons.
The regex one splits correctly as: `['I was going there...', " But now I'm here, i.e. present"]`
For my simple use cases, the regex seems to be doing OK but I will take a look at the neural one later too.

PR #173, opened 2021-07-12
NED Refactoring

Mehrad0711:
This PR refactors and simplifies bootleg code. Also introduces the following changes:
- Adding Option to add qids to text as well useful for wikidata experiments.
- Updating `transform_dataset.py` to process and filter bootleg features for paraphrasing.
- Removing `*_plus_types` fields from examples and overwriting the main fields directly.
- Creating a class for Bootleg Annotator used in server/ predict mode to annotate examples on the fly

### DIFF
--- a/genienlp/arguments.py
+++ b/genienlp/arguments.py
@@ -558,16 +558,6 @@ def check_and_update_generation_args(args):
 
 
 def post_parse_general(args):
-
-    if args.add_types_to_text != 'no' and args.add_qids_to_text != 'no' and args.add_types_to_text != args.add_qids_to_text:
-        raise ValueError('Method for adding types and qids should be the same')
-
-    for feat in args.ned_features:
-        if feat not in VALID_FEATURE_FIELDS:
-            raise ValueError(
-                'Feature {} is not supported. Please provide valid features from {} list'.format(feat, VALID_FEATURE_FIELDS)
-            )
-
     if args.val_task_names is None:
         args.val_task_names = []
         for t in args.train_task_names:
@@ -588,9 +578,6 @@ def post_parse_general(args):
             'Your val_batch_size should be divisible by number of eval_src_languages when using sentence batching.'
         )
 
-    if len(args.ned_features) != len(args.ned_features_size):
-        raise ValueError('You should specify max feature size for each feature you provided')
-
     if len(args.train_task_names) > 1:
         if args.train_iterations is None:
             args.train_iterations = [1]
@@ -599,9 +586,6 @@ def post_parse_general(args):
         if len(args.train_batch_tokens) < len(args.train_task_names):
 
             args.train_batch_tokens = len(args.train_task_names) * args.train_batch_tokens
-
-    if len(args.val_batch_size) < len(args.val_task_names):
-        args.val_batch_size = len(args.val_task_names) * args.val_batch_size
 
     # postprocess arguments
     if args.commit:
@@ -618,8 +602,6 @@ def post_parse_general(args):
     for x in ['data', 'save', 'log_dir', 'dist_sync_file']:
         setattr(args, x, os.path.join(args.root, getattr(args, x)))
 
-    args.num_features = len(args.ned_features)
-
     # tasks with the same name share the same task object
     train_tasks_dict = get_tasks(args.train_task_names, args)
     args.train_tasks = list(train_tasks_dict.values())
@@ -632,6 +614,23 @@ def post_parse_general(args):
 
 
 def post_parse_train_specific(args):
+    if args.add_types_to_text != 'no' and args.add_qids_to_text != 'no' and args.add_types_to_text != args.add_qids_to_text:
+        raise ValueError('Method for adding types and qids should be the same')
+
+    for feat in args.ned_features:
+        if feat not in VALID_FEATURE_FIELDS:
+            raise ValueError(
+                'Feature {} is not supported. Please provide valid features from {} list'.format(feat, VALID_FEATURE_FIELDS)
+            )
+
+    if len(args.ned_features) != len(args.ned_features_size):
+        raise ValueError('You should specify max feature size for each feature you provided')
+
+    args.num_features = len(args.ned_features)
+
+    if len(args.val_batch_size) < len(args.val_task_names):
+        args.val_batch_size = len(args.val_task_names) * args.val_batch_size
+
     if args.no_fast_tokenizer and args.force_fast_tokenizer:
         raise ValueError('Both no_fast_tokenizer and force_fast_tokenizer flags are on')
 
--- a/genienlp/arguments.py
+++ b/genienlp/arguments.py
@@ -487,18 +487,22 @@ def parse_argv(parser):
         '--ned_features',
         nargs='+',
         type=str,
-        default=[],
-        help='Features that will be extracted for each entity. Order is important',
+        default=['type_id', 'type_prob', 'qid'],
+        help='Features that will be extracted for each entity: "type" and "qid" are supported. Order is important',
     )
     parser.add_argument(
         '--ned_features_size',
         nargs='+',
         type=int,
-        default=[],
+        default=[1, 1, 1],
         help='Max length of each feature vector. All features are padded up to this length',
     )
     parser.add_argument(
-        '--ned_features_default_val', nargs='+', type=float, default=[], help='Default value used for each feature'
+        '--ned_features_default_val',
+        nargs='+',
+        type=float,
+        default=[0, 1.0, 0],
+        help='Max length of each feature vector. All features are padded up to this length',
     )
 
     # translation args
--- a/genienlp/arguments.py
+++ b/genienlp/arguments.py
@@ -480,7 +480,6 @@ def parse_argv(parser):
         'ngrams: lookup all ngrams in the text and see if there is a match',
     )
 
-    parser.add_argument('--verbose', action='store_true', help='Print detected types for each token')
     parser.add_argument(
         '--almond_domains', nargs='+', default=[], help='Domains used for almond dataset; e.g. music, books, ...'
     )
@@ -555,6 +554,9 @@ def check_and_update_generation_args(args):
 
 
 def post_parse_general(args):
+
+    if args.add_types_to_text != 'no' and args.add_qids_to_text != 'no' and args.add_types_to_text != args.add_qids_to_text:
+        raise ValueError('Method for adding types and qids should be the same')
 
     for feat in args.ned_features:
         if feat not in VALID_FEATURE_FIELDS:
--- a/genienlp/arguments.py
+++ b/genienlp/arguments.py
@@ -454,6 +454,14 @@ def parse_argv(parser):
         choices=['no', 'insert', 'append'],
         help='Method for adding types to input text in text-based NER approach',
     )
+
+    parser.add_argument(
+        "--add_qids_to_text",
+        default='no',
+        choices=['no', 'insert', 'append'],
+        help='Method for adding qids to input text in text-based NER approach',
+    )
+
     parser.add_argument("--ned_dump_entity_type_pairs", action='store_true', help='Dump entity type pairs')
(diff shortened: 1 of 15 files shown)


genienlp/arguments.py
```diff
@@ -472,26 +480,29 @@ def parse_argv(parser):
...
         '--almond_domains', nargs='+', default=[], help='Domains used for almond dataset; e.g. music, books, ...'
     )
     parser.add_argument(
         '--ned_features',
         nargs='+',
         type=str,
-        default=[],
-        help='Features that will be extracted for each entity. Order is important',
+        default=['type_id', 'type_prob', 'qid'],
+        help='Features that will be extracted for each entity: "type" and "qid" are supported. Order is important',
```
gcampax, line 491: We're changing the default here: what's the impact for current training configurations with and without Bootleg?

genienlp/arguments.py
```diff
@@ -472,26 +480,29 @@ def parse_argv(parser):
...
     )
     parser.add_argument(
         '--ned_features_size',
         nargs='+',
         type=int,
-        default=[],
+        default=[1, 1, 1],
         help='Max length of each feature vector. All features are padded up to this length',
```
gcampax, line 498: One part I still don't understand after all this time is the meaning of "feature vector". You have three features, and each feature is a sequence of words/classes coming out of Bootleg. What does "1" mean? What does it mean that the default value is 0 or 1.0?

genienlp/calibrate.py
```diff
@@ -44,6 +44,58 @@
...
+
+    # Options to normalize scores for a given threshold
+    parser.add_argument(
+        '--threshold',
+        type=float,
+        default=None,
+        help='The threshold above (below) which scores are considered to be positive (negative).',
+    )
+    parser.add_argument('--precision', type=float, default=None, help='Set this if scores should be normalize by precision.')
+    parser.add_argument('--recall', type=float, default=None, help='Set this if scores should be normalize by recall.')
```
gcampax, line 96: Why did you move this around? Is there any functional change here? It's hard to tell.

genienlp/data_utils/bootleg.py
```diff
@@ -33,108 +33,21 @@
...
-        model_name=args.bootleg_model,
-        verbose=False,
-    )
-    # collect all outputs now; we will filter later
-    bootleg_annotator.set_threshold(0.0)
-    setattr(bootleg_annotator, 'bootleg', bootleg)
-    return bootleg_annotator
-
-
 class Bootleg(object):
```
gcampax, line 46: So now we have "Bootleg", "BootlegAnnotator", and "BootlegAnnotator renamed to BootlegClass". This is beyond confusing. Can we pick better names, or add doc comments to distinguish?

genienlp/data_utils/bootleg.py
```diff
@@ -33,108 +33,21 @@
...
 from bootleg.run import run_model
 from bootleg.utils.parser.parser_utils import parse_boot_and_emm_args
 
-from .database_utils import is_banned
+from .database_utils import is_banned, reverse_bisect_left
 
 logger = logging.getLogger(__name__)
 
 
-def reverse_bisect_left(a, x, lo=None, hi=None):
```
gcampax, line 46: Why is this function not just called "binary_search"? What's with the comment about keeping it reverse-sorted? The function doesn't modify the input at all.

genienlp/paraphrase/scripts/transform_dataset.py
```diff
@@ -129,6 +131,9 @@ def main(args):
     if args.remove_duplicates and args.no_duplication_columns is None:
         raise ValueError('You should specify columns that define duplication')
 
+    if args.bootleg_output_file:
+        bootleg_output_file = open(args.bootleg_output_file, 'w')
```
gcampax, line 139: Given you load the whole input file in memory anyway, you should load the input file before you open the output file. That way, the same file can be both input and output.

genienlp/predict.py
```diff
@@ -512,10 +511,14 @@ def main(args):
         logger.info(f'Independent multi-GPU generation on following devices: {devices}')
         all_processes = []
         all_data_folders = split_folder_on_disk(args.data, len(devices))
+        if args.do_ned and args.ned_retrieve_method == 'bootleg':
+            all_bootleg_data_folders = split_folder_on_disk(args.bootleg_output_dir, len(devices))
```
gcampax, line 515: Why do we need this bootleg step? For train/dev, it's been run before, and for test, it will be run on the fly, right?

genienlp/run_bootleg.py
```diff
@@ -83,11 +83,12 @@ def parse_argv(parser):
...
-        '--train_tasks', nargs='+', type=str, dest='train_task_names', help='tasks to use for training', required=True
-    )
-
-    parser.add_argument(
-        '--val_tasks', nargs='+', type=str, dest='val_task_names', help='tasks to collect evaluation metrics for'
+        '--tasks',
+        nargs='+',
+        type=str,
+        dest='train_task_names',
+        help='tasks to use for bootleg; we set dest to be train_task_names so we can resue previous code for arguments processsing',
```
gcampax: "we set dest to be train_task_names so we can resue previous code for arguments processsing" has nothing to do with the help for the command-line argument. Put it in a comment.

genienlp/arguments.py
gcampax: Also, what happens if the ned_ parameters given to `train` don't match those given to `run-bootleg`?

genienlp/tasks/almond_task.py
```diff
@@ -455,16 +383,20 @@ def preprocess_field(self, sentence, field_name=None, answer=None, example_id=No
...
             ]
-        if 'type_prob' in self.args.ned_features and field_name != 'answer':
+        if self.args.do_ned and 'type_prob' in self.args.ned_features and field_name != 'answer':
             tokens_type_probs = [
                 [self.args.ned_features_default_val[1]] * self.args.ned_features_size[1] for _ in range(new_sentence_length)
             ]
+        if self.args.do_ned and 'qid' in self.args.ned_features and field_name != 'answer':
+            token_qids = [
+                [self.args.ned_features_default_val[2]] * self.args.ned_features_size[2] for _ in range(new_sentence_length)
+            ]
```
gcampax, line 399: I just noticed - why is this code in the Almond task and not in the general data loading code? How do we use NED for non-Almond tasks?

genienlp/util.py
```diff
@@ -849,7 +853,7 @@ def load_config_json(args):
                 setattr(args, r, [0])
             elif r in ['ned_features', 'ned_features_size', 'ned_features_default_val', 'override_valid_metrics']:
                 setattr(args, r, [])
-            elif r == 'add_types_to_text':
+            elif r in ['add_types_to_text', 'add_qids_to_text']:
```
gcampax: Use a tuple here, it's more idiomatic.

gcampax (commented):
Overall, I like the new design and I think it's a good, long needed cleanup. I have some questions and some comments though.

s-jse (approved):
Looks good to me, after you address @gcampax's comments.

genienlp/calibrate.py
Mehrad0711: No functional changes. I just moved the `parse_argv` to the top for predict, server, and calibrate to be consistent with the rest of the code.

gcampax: Next time, it will be helpful to have a dedicated commit for this change (with the magic words "No functional changes") so I won't worry about checking line by line.

Mehrad0711: I had this atomic commit:  "Move parse_argv method to the top consistent with the rest of the code" `10b9d9f0bb0252a8a51ccc026742b354b449785f` 
Good point on using magic words! Will do next time.

gcampax: Uh I swear I read through each commit. My bad, sorry!

Mehrad0711: No worries!

genienlp/arguments.py
Mehrad0711: For experiments that don't use ned (`do_ned` is False) there's no impact. 
For NED experiments, all features are processed by default. Now, If you want to use the features you would call `add_types_to_text` and/ or `add_qid_to_text`. I'm working on second cleanup for the code that we'll drop the `add_*_to_args` and add an argument to tell if we are adding features as text to the input or embedding ids. And then whatever features is passed as `ned_features` will be used automatically.

genienlp/arguments.py
Mehrad0711: Good catch. Since I moved out bootleg entirely out of almond task now, we don't need these arguments in `run_bootleg` anymore. I will remove them in this PR.

genienlp/data_utils/bootleg.py
Mehrad0711: Our `BootlegAnnotator` class is a wrapper for bootleg's `BootlegAnnotator` class. I can rename theirs to annotator in the import and add a docstring for our `Bootleg` and `BootlegAnnotator` classes.

gcampax: What's the purpose of our BootlegAnnotator, and what's the purpose of the Bootleg class? Maybe we can find three distinct non-confusing names.

genienlp/tasks/almond_task.py
Mehrad0711: I'm gradually moving all the code for NED out of almond task. Bootleg is now completely task-independent and can be used for other tasks too. The remaining code is for old approaches that use ES/ local database for NER. I will be moving these out also in the next PRs.

genienlp/paraphrase/scripts/transform_dataset.py
Mehrad0711: Sure.

genienlp/predict.py
Mehrad0711: This is needed for the paraphrase filtering step which is done on 4 gpus. The prepped bootleg files need to split the same way that input paraphrases are split.

genienlp/data_utils/bootleg.py
Mehrad0711: `Bootleg` class is a wrapper for all the functionalities we need from bootleg. It calls the appropriate functions from bootleg, processes the collected results, extract the needed features, and add those features to our examples.

`BootlegAnnotator` does the same but in a on-the-fly fashion useful for server mode when input data is unknown ahead of time.

I kinda like the current names but if the conflict is with the original `BootlegAnnotator` we can rename ours to `BootlegAnnotatorWrapper`.

genienlp/util.py
Mehrad0711: Fixed

genienlp/run_bootleg.py
Mehrad0711: Fixed.

genienlp/data_utils/bootleg.py
Mehrad0711: This is binary_search for a reverse sorted list (i.e. descending order). I adopted the naming convention in `bisect` library. 
Yeah, the comment is misphrased. It should be: "Locate the insertion point for x in a to **maintain** its reverse sorted order"

Mehrad0711:
I pushed a hotfix for bootleg to master and rebased this branch. That pushed all the previous commits ahead of your comments. The only new changes after your reviews are in the last two commits.

genienlp/arguments.py
Mehrad0711: The features are defined per token and include information about the wikidata qids, types, and probs assigned by bootleg . Each token can have up to 3 features defined by `ned_features`. Each of these features is a list of numbers. The size and default value for this list is defined by `ned_features_size` and `ned_features_default_val`.
For example, if size for type_id feature is 1, we only keep the first type associated with an entity Qid. 
If size for qid is 2, we keep the top 2 qids predicted by bootleg.
The default value is used to pad features up to the maximum size.

PR #176, opened 2021-07-22
Add sequence classifier

kevintangzero:


Mehrad0711:
Thanks for your work @kevintangzero!
Can you please add some tests for the new sentence classification task? You can look at [token classification tests](https://github.com/stanford-oval/genienlp/blob/master/tests/test_token_classification.sh) as an example.

genienlp/tasks/generic_dataset.py
```diff
@@ -1929,3 +1929,52 @@ def return_splits(cls, name, path='.data', train='train', validation='dev', test
...
+        else:
+            for label in ['pos', 'neg']:
+                for fname in glob.iglob(os.path.join(path, label, '*.tsv')):
+                    with open(fname, 'r') as f:
+                        line = f.readline()
+                        while line:
+                            context = line.split('\t')[2]
+                            answer = labels[label]
+                            examples.append(Example.from_raw(
+                                make_example_id(self, len(examples)),
```
gcampax: You actually have the example ID in the file already, you don't need to construct it dynamically. This construction seems to depend on the globbing order for files which might not be reliable.

genienlp/models/transformer_sequence_classification.py
```diff
@@ -0,0 +1,104 @@
...
+        if self.training:
+            batch = input[0]
+            outputs = self.model(
+                batch.context.value,
+                labels=batch.answer.value,
+                attention_mask=(batch.context.value != self.numericalizer.pad_id),
+            )
+            return outputs
+        else:
+            return self.model(**kwargs)
```
gcampax, line 102: I'm not sure this if-else is correct.
The training part looks right, but for the inference part I think you need to write a new branch in [generate_with_model](https://github.com/stanford-oval/genienlp/blob/master/genienlp/validate.py#L42) and make sure whatever arguments you pass there are passed here.

Maybe @Mehrad0711 can confirm here.

gcampax (commented):
I think this looks good so far. I'm not sure about the generate_with_model thing.

Did you run this code at all? Did you try training a model?

genienlp/models/transformer_sequence_classification.py
Mehrad0711: The model code looks correct but [this if statement](https://github.com/stanford-oval/genienlp/blob/6e45a5699d4491975a2385930383d26ecd9eb321/genienlp/validate.py#L55) should include `TransformerForSeqeunceClassification` as well.  I think the `generate_with_classification_model` should work for sequence classification task as well, but we need to run the code and make sure that's the case.

genienlp/tasks/generic_dataset.py
```diff
@@ -1929,3 +1929,50 @@ def return_splits(cls, name, path='.data', train='train', validation='dev', test
...
+            logger.info(f'Loading cached data from {cache_name}')
+            examples = torch.load(cache_name)
+        else:
+            for label in ['pos', 'neg']:
+                for fname in glob.iglob(os.path.join(path, label, '*.tsv')):
+                    with open(fname, 'r') as f:
+                        line = f.readline()
+                        while line:
+                            context = line.split('\t')[2]
+                            answer = labels[label]
```
Mehrad0711: The answer is supposed to be a digit wrapped as a string so the labels mapping should be:
`labels = {'neg': '0', 'pos': '1'}`
Remember the classification model outputs a digit.

genienlp/tasks/generic_dataset.py
```diff
@@ -1929,3 +1929,50 @@ def return_splits(cls, name, path='.data', train='train', validation='dev', test
...
+        examples = []
+        labels = {'neg': 'negative', 'pos': 'positive'}
+        question = 'Is this sentence in-domain or out-domain?'
+
+        cache_name = os.path.join(cached_path, os.path.basename(path))
+        if os.path.exists(cache_name) and not skip_cache:
+            logger.info(f'Loading cached data from {cache_name}')
+            examples = torch.load(cache_name)
+        else:
+            for label in ['pos', 'neg']:
```
Mehrad0711: Not sure why we need two separate files. The last column of your files contains this information (it's pos if it has thingtalk, otherwise it's ood)

genienlp/validate.py
```diff
@@ -245,7 +245,9 @@ def generate_with_classification_model(
         labels = batch.answer.value.tolist()
 
         logits = output.logits
-        predictions = torch.argmax(logits, dim=2).tolist()
+        predictions = torch.argmax(logits, dim=-1).tolist()
```
gcampax, line 248: Can we add a comment here explaining the shapes in the different situations?

genienlp/data_utils/example.py
```diff
@@ -172,6 +172,33 @@ def from_examples(examples: Iterable[Example], numericalizer):
...
+                    batch_decoder_numerical.append(numericalizer.decoder_vocab.encode(answers[i]))
+            else:
+                batch_decoder_numerical = [[]] * len(answers)
+
+            tokenized_answers = []
+            for i in range(len(answers)):
+                tokenized_answers.append(
+                    SequentialField(
+                        value=answers[i],
+                        length=len(answers[i]),
```
gcampax, line 196: `len(answers[i])` is always 1 here right?

genienlp/data_utils/example.py
```diff
@@ -172,6 +172,33 @@ def from_examples(examples: Iterable[Example], numericalizer):
...
+                ]
+                for ex in examples
+            ]
+
+            batch_decoder_numerical = []
+            if numericalizer.decoder_vocab:
+                for i in range(len(answers)):
+                    batch_decoder_numerical.append(numericalizer.decoder_vocab.encode(answers[i]))
+            else:
+                batch_decoder_numerical = [[]] * len(answers)
```
gcampax, line 189: batch_decoder_numerical doesn't make much sense for a sequence_classification task, it is meant for seq2seq models that use a different output vocabulary.

genienlp/tasks/generic_task.py
```diff
@@ -378,3 +378,18 @@ def _make_example(self, parts, dir_name=None, **kwargs):
...
+@register_task('ood_task')
+class OODTask(BaseTask):
+    def __init__(self, name, args):
+        self.id2label = ['0', '1']
+        self.num_labels = 2
+        super().__init__(name, args)
+
+    @property
+    def metrics(self):
+        return ['recall', 'precision']
```
gcampax: The first metric is the one we use to choose the best model during training, and should probably be F1

genienlp/tasks/generic_dataset.py
```diff
@@ -1929,3 +1929,48 @@ def return_splits(cls, name, path='.data', train='train', validation='dev', test
...
+    def __init__(self, path, lower=False, cached_path=None, skip_cache=False, **kwargs):
+        examples = []
+        question = 'Is this sentence in-domain or out-domain?'
+
+        cache_name = os.path.join(cached_path, os.path.basename(path))
+        if os.path.exists(cache_name) and not skip_cache:
+            logger.info(f'Loading cached data from {cache_name}')
+            examples = torch.load(cache_name)
+        else:
+            for fname in glob.iglob(os.path.join(path, '*.tsv')):
```
gcampax: What is the on-disk format now? Is it different directories for train/eval/test? Do we call it "dev" or "eval"?

gcampax (approved):
I think this looks good. Let's merge it and get it integrated with the rest of the pipeline.

tests/test_sequence_classification.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+. ./tests/lib.sh
+
+genienlp train --train_tasks ood_task --model TransformerForSequenceClassification --pretrained_model distilbert-base-uncased --save $workdir/model --train_iterations 4 --save_every 2 --log_every 2 --val_every 2 --data $SRCDIR/dataset/ood_task/ --skip_cache --force_fast_tokenizer --train_batch_tokens 200
```
Mehrad0711: Can you please add a test for prediction too and checking against expected results. Again you can look at other tests for how it's done. It will ensure the later changes we make to the code do not change the output unexpectedly.

genienlp/metrics.py
```diff
@@ -210,6 +210,28 @@ def lower(text):
...
+    prediction_tokens = prediction.split()
+    ground_truth_tokens = ground_truth.split()
+    common = collections.Counter(prediction_tokens) & collections.Counter(ground_truth_tokens)
+    num_same = sum(common.values())
+    if num_same == 0:
+        return 0
+    recall = 1.0 * num_same / len(ground_truth_tokens)
+    return recall
+
+
```
Mehrad0711: Do you still need this code now that you are using HF metrics?

tests/test_sequence_classification.sh
```diff
@@ -1,5 +1,4 @@
 #!/usr/bin/env bash
 
 . ./tests/lib.sh
-
```
Mehrad0711: 100 `train_iterations` is unnecessary and will slow down the test. To get meaningful predictions ~20 iterations should be enough. Also `save_every`, `log_every`, and `val_every` can be set to half of `train_iterations` for this test.

Mehrad0711 (approved):
Looks good now.

Issue #175, opened 2021-07-15
Bootleg type mapping should be configurable


gcampax:
Currently, the mapping between Almond types and Bootleg QIDs is defined in a fixed JSON file (es_material/almond_type2qid.json) which also depends on the domain.

Instead, the path to this file should be configurable so this file can be dynamically generated during synthesis based on the information in Thingpedia manifests.

Filing this issue for tracking so we don't forget.

gcampax:
After discussion on Slack, we agreed the NED code needs some refactoring:

We agreed to:
1. [x] we rename "Feature" to "Entity"
2. [x] we stop using the word QID everywhere, and either say "type_id" or "entity_id" so we don't get confused everywhere
3. [x] we make type_id to be the (numeric) QID of the type given by bootleg - or a "normalized" type QID given by the almond_type_* files, which we simplify to be "exact" and "substring" in one file, or maybe some glob-like thing and a simple mapping
4. [x] we remove "ned_features", "ned_features_default", and "ned_feature_size", and replace it with "max_entity_per_span" or similar which is just a number, tells you how many things to put with |
5. [x] we make a "add_entity_to_text" = (type, qid) to decide what to append to the entity
6. [x] we put the wikidata type name of the (normalized) type id in the input

Mehrad0711:
Addressed in #183

PR #183, opened 2021-08-01
NED Refactoring (2)

Mehrad0711:
Addressing https://github.com/stanford-oval/genienlp/issues/175


genienlp/arguments.py
```diff
@@ -450,17 +457,25 @@ def parse_argv(parser):
...
     )
 
     parser.add_argument(
-        "--add_qids_to_text",
-        default='no',
-        choices=['no', 'insert', 'append'],
-        help='Method for adding qids to input text in text-based NER approach',
+        "--entity_attributes",
+        nargs='+',
+        default=['type_id', 'type_prob', 'qid'],
```
gcampax: What does it mean to set entity_attributes to "type_prob"? Why is this not just "type" and "qid"?

genienlp/data_utils/bootleg.py
```diff
@@ -63,35 +63,36 @@ def __init__(self, args):
...
-
-        self.almond_type2qid = dict()
-        with open(f'{self.args.database_dir}/es_material/almond_type2qid.json', 'r') as fin:
-            almond_type2qid = ujson.load(fin)
-        for domain in self.almond_domains:
-            self.almond_type2qid.update(almond_type2qid[domain])
-        self.qid2almond_type = {v: k for k, v in self.almond_type2qid.items()}
-
+        self.almond_type_mapping = dict()
+        self.wiki2normalized_type = dict()
```
gcampax: Add a couple comments describing what are the keys and what are the values here.

genienlp/data_utils/bootleg.py
```diff
@@ -63,35 +63,36 @@ def __init__(self, args):
...
+
+        if getattr(self.args, 'almond_type_mapping_path', None):
+            with open(os.path.join(self.args.root, self.args.almond_type_mapping_path)) as fin:
+                self.almond_type_mapping = ujson.load(fin)
+            for normalized_type, titles in self.almond_type_mapping.items():
+                for title in titles:
+                    self.wiki2normalized_type[title] = normalized_type
```
gcampax: Maybe factor this snippet of code into a method that gets the filename and loads it?

genienlp/data_utils/bootleg.py
```diff
@@ -177,19 +178,16 @@ def disambiguate_mentions(self, config_args):
...
-                    types = self.type_mapping_include[key]
+        for key in self.wiki2normalized_type.keys():
+            if fnmatch.fnmatch(title, key):
+                types = self.wiki2normalized_type[key]
 
         typeqids = None
         if types is not None:
             if isinstance(types, str):
-                typeqids = [self.almond_type2qid[types]]
+                typeqids = [self.type_vocab_to_entityqid[types]]
```
gcampax: _to_entityqid? You mean to typeqid here?

genienlp/data_utils/bootleg.py
```diff
@@ -177,19 +178,16 @@ def disambiguate_mentions(self, config_args):
 
     def post_process_bootleg_types(self, title):
         types = None
-        if title in self.type_mapping_match:
-            types = self.type_mapping_match[title]
-        else:
-            for key in self.type_mapping_include.keys():
-                if key in title:
-                    types = self.type_mapping_include[key]
+        for key in self.wiki2normalized_type.keys():
```
gcampax: It looks like wiki2normalized_type is never actually used as a dictionary. Should we make it a list of pairs instead?

genienlp/data_utils/bootleg.py
```diff
@@ -198,35 +196,35 @@ def add_type_tokens(self, sentence, features):
...
+                if any([val != 0 for val in feat.type_id]):
                     final_token = '<e> '
                     final_types = ''
-                    if self.args.add_types_to_text != 'no':
+                    if 'type_id' in self.args.entity_attributes:
                         if self.bootleg_post_process_types:
                             all_types = ' | '.join(
-                                set(
-                                    self.qid2almond_type[self.id2typeqid[id]]
+                                sorted(
```
gcampax: Do you need both "set" and "sorted" to remove duplicates?

genienlp/data_utils/bootleg.py
```diff
@@ -198,35 +196,35 @@ def add_type_tokens(self, sentence, features):
...
                                     if id != self.args.db_unk_id
                                 )
                             )
                         final_types = '( ' + all_types + ' ) '
                     final_qids = ''
-                    if self.args.add_qids_to_text != 'no':
-                        all_qids = ' | '.join(set('Q' + str(id) for id in feat.qid))
+                    if 'qid' in self.args.entity_attributes:
+                        all_qids = ' | '.join(sorted('Q' + str(id) for id in feat.qid if id != -1))
                         final_qids = '[' + all_qids + ']'
```
gcampax: I think you should extract all the code that computes the string to append (with types/qids) into a method that takes the Entity object. The same code is used in insert or append mode.

genienlp/data_utils/database.py
```diff
@@ -38,25 +38,24 @@
 
 
 class Database(object):
```
gcampax, line 37: This class is unused right? It's pre-bootleg and we can kill it. Even your NER experiments will be Bootleg-only.

genienlp/data_utils/example.py
```diff
@@ -64,19 +69,18 @@ def flatten(self):
...
-    pad_feature = None
-    if len(feature_fields):
-        pad_feature = Feature()
-        for i, field in enumerate(feature_fields):
-            assert field in VALID_FEATURE_FIELDS
-            setattr(pad_feature, field, [ned_features_default_val[i]] * ned_features_size[i])
-    return pad_feature
+# 1: to remove self
+signature = inspect.signature(Entity.__init__)
+VALID_FEATURE_FIELDS = tuple(signature.parameters.keys())[1:]
```
gcampax: Honestly, this would be more readable by writing explicitly `["type_id", "type_prob", "qid"]`. It's three fields, and defined right above it.

genienlp/data_utils/example.py
```diff
@@ -160,7 +164,7 @@ def from_examples(examples: Iterable[Example], numericalizer):
             all_context_plus_question_features.append(context_plus_question_feature)
 
         features = None
-        if args.do_ned and args.add_types_to_text == 'no' and args.add_qids_to_text == 'no':
+        if args.do_ned and args.add_entities_to_text == 'no':
             features = [a for a in all_context_plus_question_features if a]
```
gcampax: When is a Entity false-y? It's an object and it's always set, right? At best, it would be empty (pad entity) if there are no entities for a given token

genienlp/data_utils/remote_database.py
```diff
@@ -54,14 +54,13 @@ def __getattr__(self, attr_name):
 
 
 class RemoteElasticDatabase(object):
```
gcampax: Kill this. We don't use it, we don't want it, it's unused complexity.

genienlp/models/identity_encoder.py
```diff
@@ -135,15 +135,15 @@ def forward(self, batch):
 
         context_entity_ids, context_entity_probs, context_entity_masking = None, None, None
 
-        if self.args.num_db_types > 0 and self.args.add_types_to_text == 'no':
-            context_entity_ids = batch.context.feature[:, :, : self.args.ned_features_size[0]].long()
+        if self.args.num_db_types > 0 and self.args.add_entities_to_text == 'no':
```
gcampax: What's "num_db_types"? Should this be "args.do_ner" instead?

genienlp/tasks/almond_task.py
```diff
@@ -96,18 +96,14 @@ def _init_db(self):
             with open(os.path.join(self.args.database_dir, 'es_material/typeqid2id.json'), 'r') as fin:
                 typeqid2id = ujson.load(fin)
 
-            self.db = Database(
-                canonical2type, typeqid2id, all_canonicals, self.args.ned_features_default_val, self.args.ned_features_size
-            )
+            self.db = Database(canonical2type, typeqid2id, all_canonicals, self.args.max_features_size)
```
gcampax: More killing with more fire, just always use Bootleg.

genienlp/tasks/almond_task.py
```diff
@@ -240,9 +236,7 @@ def pad_features(self, features, max_size, pad_id):
         return features
 
     def oracle_type_ids(self, tokens, entity2type):
```
gcampax, line 229: Do you still need the oracle? It's so hacky, I wonder if we can do it with a separate script that generates oracle bootleg_features.jsonl

genienlp/tasks/almond_task.py
```diff
@@ -386,26 +376,18 @@ def preprocess_field(self, sentence, field_name=None, answer=None, example_id=No
...
-            ]
-        if self.args.do_ned and 'qid' in self.args.ned_features and field_name != 'answer':
-            token_qids = [
-                [self.args.ned_features_default_val[2]] * self.args.ned_features_size[2] for _ in range(new_sentence_length)
-            ]
+        if self.args.do_ned and field_name != 'answer':
+            tokens_type_ids = [[0] * self.args.max_features_size for _ in range(new_sentence_length)]
+        if self.args.do_ned and field_name != 'answer':
+            tokens_type_probs = [[0] * self.args.max_features_size for _ in range(new_sentence_length)]
+        if self.args.do_ned and field_name != 'answer':
```
gcampax: Same if statement repeated three times?

genienlp/util.py
```diff
@@ -785,9 +787,8 @@ def load_config_json(args):
             'ned_retrieve_method',
             'database_lookup_method',
             'almond_domains',
-            'ned_features',
-            'ned_features_size',
-            'ned_features_default_val',
+            'almond_type_mapping_path',
+            'max_features_size',
```
gcampax, line 791: Do we need compatibility with the old option names in config.json? Otherwise existing trained models will start failing because we won't apply Bootleg correctly.

genienlp/data_utils/bootleg.py
```diff
@@ -63,35 +63,36 @@ def __init__(self, args):
...
+            for normalized_type, titles in self.almond_type_mapping.items():
+                for title in titles:
+                    self.wiki2normalized_type[title] = normalized_type
+        else:
+            with open(
+                os.path.join(os.path.dirname(os.path.abspath(__file__)), 'database_files/almond_type_mapping.json')
+            ) as fin:
+                almond_type_mapping_all_domains = ujson.load(fin)
+            # only keep subset for provided domains
+            for domain in self.almond_domains:
```
gcampax: Isn't the flag called --ner_domains ?

gcampax (commented):
Its getting there but there is more to do.

genienlp/arguments.py
Mehrad0711: 'type_prob' and 'type_id' always go together for bootleg cause probability is used for filtering. For embedding-based ned approach `type_prob` is used to weight the type embeddings so it's optional.  
Once we remove the embedding based approach we can remove `type_prob`. For text based approach you can just pass `type_id` or we can set the default here to be `type_id` since we don't use `qid` except in wikidata experiments.

gcampax: Yes the default should be type_id. Do you really need to ablate on type_id without type_prob when you do the embedding experiment?

genienlp/data_utils/bootleg.py
Mehrad0711: `set()` is not useful here since the additional types are already omitted at this point. The set behavior is here: https://github.com/stanford-oval/genienlp/blob/49ab643afd608a60de22b27e804d3f9d215f91f6/genienlp/data_utils/bootleg.py#L345
to avoid appending similar types before trimming the extra ones.

genienlp/data_utils/bootleg.py
Mehrad0711: Is there an advantage to lists? dictionary seems more consistent since all the other mappings are a dictinary

genienlp/data_utils/database.py
Mehrad0711: I will kill `RemoteDatabase` but need this one as a baseline against bootleg.

genienlp/data_utils/bootleg.py
gcampax: There is a marginal speed advantage. Also, it would preserve the ordering given in the json file in case there are overlapping patterns.

genienlp/data_utils/example.py
Mehrad0711: I remember it was possible before for question and context features to be empty but I can't find that code now. Perhaps it got refactored. I'll remove it then.

genienlp/tasks/almond_task.py
Mehrad0711: That's a good idea but it needs some additional work. I'll revisit it later if we decide to keep it after submission.

genienlp/arguments.py
Mehrad0711: Yes probably

genienlp/util.py
Mehrad0711: I can add sth that calculates `max_features_size` from `ned_features_size ` and maps the old args to new args but not sure if it would give the exact same results for old models cause the way we keep qids and the types per qid is now configurable as opposed to before.
I imagine we can use the commit hash stored in the config to determine the genienlp version for reevaluating old models?

genienlp/arguments.py
```diff
@@ -496,9 +496,7 @@ def parse_argv(parser):
         'ngrams: lookup all ngrams in the text and see if there is a match',
     )
 
-    parser.add_argument(
-        '--almond_domains', nargs='+', default=[], help='Domains used for almond dataset; e.g. music, books, ...'
-    )
+    parser.add_argument('--ned_domains', nargs='+', default=[], help='Domains used for almond dataset; e.g. music, books, ...')
```
gcampax, line 496: You still have ned_domains and ner_domains? Is that necessary?

genienlp/data_utils/bootleg.py
```diff
@@ -141,6 +122,34 @@ def __init__(self, args):
...
+        # a list of tuples: each pair includes a wiki type and their normalized type
+        self.wiki2normalized_type = list()
+    
+        if self.args.almond_type_mapping_path:
+            # read mapping from user-provided file
+            with open(os.path.join(self.args.root, self.args.almond_type_mapping_path)) as fin:
+                self.almond_type_mapping = ujson.load(fin)
+            for normalized_type, titles in self.almond_type_mapping.items():
+                for title in titles:
+                    self.wiki2normalized_type.append((title, normalized_type))
```
gcampax: My suggestion was to take the parts common to the two branches of the if statement and make them into a method. Moving the whole thing into a method doesn't help (as much): you still have the same complexity in terms of nested control flow and redundant code.

genienlp/data_utils/bootleg.py
```diff
@@ -178,19 +187,45 @@ def disambiguate_mentions(self, config_args):
...
-                typeqids = [self.type_vocab_to_entityqid[types]]
+                typeqids = [self.type_vocab_to_typeqid[types]]
             elif isinstance(types, (list, tuple)):
-                typeqids = [self.type_vocab_to_entityqid[type_] for type_ in types]
+                typeqids = [self.type_vocab_to_typeqid[type_] for type_ in types]
 
         return typeqids
-
+    
+    def textualize_features(self, feat):
```
gcampax: Maybe "features_to_text" ? Or better, "entities_to_text", cause we're moving away from the generic word "feature"

genienlp/data_utils/bootleg.py
```diff
@@ -178,19 +187,45 @@ def disambiguate_mentions(self, config_args):
...
+        if 'type_id' in self.args.entity_attributes:
+            if self.bootleg_post_process_types:
```
gcampax: What's the difference between these two branches? Why are there two branches at all, given that postprocessing of types occurs when entities are loaded and not here?

genienlp/data_utils/bootleg.py
```diff
@@ -246,34 +260,13 @@ def add_type_tokens(self, sentence, features):
...
-                        all_qids = ' | '.join(sorted('Q' + str(id) for id in feat.qid if id != -1))
-                        final_qids = ['[', all_qids, ']']
+                    final_types, final_qids = self.textualize_features(feat)
                     all_tokens = []
                     # append all entities with same type
                     while i < len(sentence_tokens) and features[i] == feat:
                         all_tokens.append(sentence_tokens[i])
                         i += 1
-                    final_token = ' '.join([*all_tokens, *final_types, *final_qids, ';'])
+                    final_token = ' '.join([*all_tokens, final_types, *[final_qids], ';'])
```
gcampax: What's the wrapping of final_qids doing?

genienlp/arguments.py
Mehrad0711: `ner_domains` are for CrossNER task since it has multiple domains. I can rename it to `crossner_domains`

genienlp/data_utils/bootleg.py
Mehrad0711: Only shared code between the two branches is this three line of code: https://github.com/stanford-oval/genienlp/blob/58aecfc46e70df5b2069d15292ccfd1ef9282c03/genienlp/data_utils/bootleg.py#L149-L151
Do you really want to refactor them into a method?

genienlp/data_utils/bootleg.py
Mehrad0711: I remember this branching was needed for non-bootleg NED but after this refactoring I can't tell if it's still the case.
I remove it for now.

s-jse (approved):

genienlp/data_utils/bootleg.py
gcampax: Yes because those three lines of code have two loops in them, and they nested within other control flow. Have you heard of cyclomatic complexity? It's basically a measure of how nested your if statements & for loops go, and the higher the number the less readable code is.

genienlp/arguments.py
```diff
@@ -521,7 +521,7 @@ def parse_argv(parser):
 
     # token classification task args
     parser.add_argument('--num_labels', type=int, help='num_labels for classification tasks')
-    parser.add_argument('--ner_domains', nargs='+', type=str, help='domains to use for CrossNER task')
+    parser.add_argument('--crossner_domains', nargs='+', type=str, help='domains to use for CrossNER task')
```
gcampax, line 524: I don't understand, we're going in circles. Why is this a separate argument?

Mehrad0711: Crossner dataset has multiple subfolders. Each containing train/eval/test splits per domain. This argument indicates which domain we wanna do an experiment on. This argument has nothing to do with NED.

gcampax: I feel like that should be a pipeline problem then, but ok at least it's a scoped argument specific to the task, like the almond_* stuff.

PR #186, opened 2021-08-05
Fix bug when preprocessing special tokens with T5 tokenizer

Mehrad0711:
And no functional changes for other tokenizers.


gcampax (approved):
I think this looks good. To be clear, there are no functional changes for non T5 tokenizer, correct?

Mehrad0711:
Yes. Just updated my comment.

PR #187, opened 2021-08-05
NED Refactoring (3)

Mehrad0711:
- Fixing type and entity oracle methods  for NED
- Unplugging database from the task so it can run independently similar to bootleg 
- BootlegAnnotator now inherits from Bootleg and Bootleg inherits from Database.


genienlp/data_utils/bootleg.py
```diff
@@ -373,20 +374,18 @@ def replace_features_inplace(self, examples, all_token_type_ids, all_tokens_type
         for n, (ex, tokens_type_ids, tokens_type_probs, tokens_qids) in enumerate(
             zip(examples, all_token_type_ids, all_tokens_type_probs, all_tokens_qids)
         ):
+            features = [Entity(*tup) for tup in zip(*[tokens_type_ids, tokens_type_probs, tokens_qids])]
```
gcampax: Why do you need *[ here?

genienlp/train.py
```diff
@@ -87,11 +88,13 @@ def initialize_logger(args):
...
-    bootleg = None
-    if args.do_ned and args.ned_retrieve_method == 'bootleg':
-        bootleg = Bootleg(args)
+    # initialize bootleg/ database
+    bootleg, db = None, None
+    if args.do_ned:
+        if args.ned_retrieve_method == 'bootleg':
+            bootleg = Bootleg(args)
+        else:
+            db = Database(args)
```
gcampax, line 96: I wonder if we can make Bootleg and Database have each one method with the same name and signature, so you just instantiate the right class here and handle it transparently later.

In fact, given that you changed Bootleg to extend Database, it seems natural to unify this code.

I imagine you would have:

- AbstractEntityDisambiguator method that has the common code across all NED methods, with most methods stubbed out
This would have an abstract `process_examples(examples, path, utterance_field, batch=True)` method that would be designed to modify the examples in-place (similar to the method how you have so far).
The method adds Entity objects to each token, and adds text to the input sentence.

- BootlegEntityDisambiguator extends AbstractEntityDisambiguator
Implements `process_examples` as:
  - if batch==False, lazily initialize a BootlegAnnotator, and use it to annotate each example
  - if batch==True, check for an appropriately named jsonl file of features
    - if missing, create it
    - if present, load it
 
Has method `dump_entities` that writes the JSONL file. `run-bootleg` calls that method directly.

- NaiveEntityDisambiguator extends AbstractEntityDisambiguator
Implements `process_example` by looking up examples in the database, naively.

- OracleEntityDisambiguator extends AbstractEntityDisambiguator

That way, we instantiate the entity disambiguator once (maybe even in a utility method shared across server, predict, train) and pass it around, and all the code doesn't need to care.

gcampax (changes requested):
This is a very good cleanup! But let's go further and clean up this oracle/naive stuff once and for all, so all the research code is hidden behind an interface product side doesn't need to worry about.

genienlp/data_utils/bootleg.py
Mehrad0711: I don't know why I keep doing this!

genienlp/train.py
Mehrad0711: This was going to be the next PR(!) and you laid out the details nicely. I was thinking of keeping Naive and Oracle in the same class but I think your proposal is better. 
I have a few suggestions around the naming:
- Instead of `XXXEntityDisambiguator` I was thinking of `XXXEntityLinker` cause Naive and Oracle are not disambiguators: they can't tell you the Qid. They just map (i.e. link) a mention to an entity based on some rules. 
- Instead of `batch`, I was thinking of `mode=(server | normal)`. `batch` would be a bit confusing since both Bootleg and BootlegAnnotator process examples in batches.

What do you think?

Mehrad0711: Actually, I think we should keep BootlegEntityDisambiguator and BootlegAnnotatorEntityDisambiguator as separate classes. They each would have their own `process_example` method. Otherwise, we have to cram two different codes into one and using mode for branching.

genienlp/data_utils/bootleg.py
gcampax: Should have a lint, catch all instances of *[, cause everytime I see it I get confused and think it's some obscure python thing!

genienlp/train.py
gcampax: Ok to leave them as separate classes, but let's make them `BatchBootlegEntityDisambiguator` and `StreamingBootlegEntityDisambiguator`.
I think "batch" is clear that it means "the whole thing as a one-off batch job", even if the "Streaming" version has minibatches.
Alternative is `DatasetBootlegEntityDisambiguator` but then we need a doc string to explain the name.

Mehrad0711: I'm Ok with those two. But how about `BootlegEntityDisambiguator` and `BootlegEntityDisambiguatorFor(Server|Serving)` ?

gcampax: No I don't want to have a name that is a substring of the other, because it makes it look like it's a subclass instead of a sibling class.

Mehrad0711: Ok and then how about `ServingBootlegEntityDisambiguator` instead of `StreamingBootlegEntityDisambiguator`. Pros/ Cons?

gcampax: Serving or Inference are fine with me if you don't want to use Streaming.

gcampax: Btw I don't mind EntityLinker, but I chose EntityDisambiguator because the cmdline arguments are "ned", not "nel"

gcampax:
Anyway, this is superseded by #188, and there is no reason to merge this one without the other.

PR #188, opened 2021-08-05
NED Refactoring (4)

Mehrad0711:


README.md
```diff
@@ -149,7 +149,7 @@ It will also generate several files in <results_temp> folder. In `eval_bootleg/[
 
 Now you can use the extracted features from bootleg in downstream tasks such as semantic parsing to improve named entity understanding and consequently generation:
 ```bash
-genienlp train --train_tasks <train_task_names> --train_iterations <iterations> --preserve_case --save <savedir> --data <dataset_dir> --model TransformerSeq2Seq --pretrained_model facebook/bart-base --train_batch_tokens 1000 --val_batch_size 1000 --do_ned --database_dir <database_dir> --ned_retrieve_method bootleg --entity_attributes type_id type_prob --add_entities_to_text append --bootleg_model <bootleg_model>
+genienlp train --train_tasks <train_task_names> --train_iterations <iterations> --preserve_case --save <savedir> --data <dataset_dir> --model TransformerSeq2Seq --pretrained_model facebook/bart-base --train_batch_tokens 1000 --val_batch_size 1000 --do_ned --database_dir <database_dir> --ned_class_name bootleg --entity_attributes type_id type_prob --add_entities_to_text append --bootleg_model <bootleg_model>
```
gcampax: Please undo this change, ned_retrieve_method is a much more informative argument.

genienlp/ned/ned_utils.py
```diff
@@ -281,18 +281,18 @@ def init_ned_model(args, ned_class_name=None):
     ned_model = None
     if args.do_ned:
         if ned_class_name is None:
-            if args.ned_retrieve_method == 'bootleg':
+            if args.ned_class_name == 'bootleg':
                 ned_class_name = 'Bootleg'
-            elif args.ned_retrieve_method == 'bootleg-annotator':
+            elif args.ned_class_name == 'bootleg-annotator':
```
gcampax: ned_retrieve_method can never be bootleg-annotator, right? The difference between bootleg and bootleg-annotator is whether you're called from "server" or "train" / "predict"

genienlp/ned/base.py
```diff
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2020-2021 The Board of Trustees of the Leland Stanford Junior University
+# Copyright (c) 2021-2022 The Board of Trustees of the Leland Stanford Junior University
```
gcampax: 2022?

genienlp/ned/base.py
```diff
@@ -0,0 +1,333 @@
...
+        self.unk_id = 0
+        self.unk_type = self.id2typeqid[self.unk_id]
+        self.ckpt_name = None
+
+    def update_wiki2normalized_type(self):
+        for normalized_type, titles in self.almond_type_mapping.items():
+            for title in titles:
+                self.wiki2normalized_type.append((title, normalized_type))
+
+    def process_examples(self, examples, split_path, utterance_field):
```
gcampax: How can this method have an implementation if this is an abstract class?

genienlp/ned/base.py
```diff
@@ -0,0 +1,333 @@
...
+                continue
+            ent_num_tokens = len(ent.split(' '))
+            idx = tokens_text.index(ent)
+            token_pos = len(tokens_text[:idx].strip().split(' '))
+            type = self.typeqid2id.get(self.alias2type[ent], self.unk_id)
+            tokens_type_ids[token_pos : token_pos + ent_num_tokens] = [[type] * self.max_features_size] * ent_num_tokens
+
+        return tokens_type_ids
+
+    def lookup(self, tokens, database_lookup_method=None, min_entity_len=2, max_entity_len=4):
```
gcampax: This method is not used by Bootleg right? So it doesn't belong here. It should go into a subclass.

genienlp/ned/bootleg.py
```diff
@@ -30,80 +30,44 @@
...
-        self.cur_entity_embed_size = 0
-
-        # Mapping between model directory and checkpoint name
-        model2checkpoint = {'bootleg_uncased_mini': 'bootleg_wiki.pth', 'bootleg_uncased_super_mini': 'bootleg_wiki.pth'}
-
-        self.ckpt_name, extension = model2checkpoint[self.args.bootleg_model].split('.')
-        self.model_ckpt_path = os.path.join(self.model_dir, self.ckpt_name + '.' + extension)
+        ###
+
+        self.ckpt_name = 'bootleg_wiki'
```
gcampax, line 69: This is a cmdline argument right?

genienlp/ned/bootleg.py
```diff
@@ -173,21 +132,16 @@ def extract_mentions(self, input_path):
...
-        if len(tokens) > max_size:
-            tokens = tokens[:max_size]
-        else:
-            tokens += [pad_id] * (max_size - len(tokens))
-        return tokens
-
     def disambiguate_mentions(self, config_args):
         run_model(self.args.bootleg_dump_mode, config_args)
 
     def post_process_bootleg_types(self, title):
```
gcampax, line 138: Why is the initialization for type postprocessing done in the parent class, but type postprocessing is only used in Bootleg?

genienlp/ned/bootleg.py
```diff
@@ -323,103 +215,70 @@ def collect_features_per_line(self, line, threshold):
...
+        self.create_jsonl(path, examples, utterance_field)
+
+        # extract mentions and mention spans in the sentence and write them to output jsonl files
+        self.extract_mentions(path)
+
+        # find the right entity candidate for each mention
+        self.disambiguate_mentions(config_args)
+
+
+class ServingBootlegEntityDisambiguator(BatchBootlegEntityDisambiguator):
```
gcampax, line 279: Do you need this inheritance? What are you inheriting?

genienlp/ned/main.py
```diff
@@ -0,0 +1,190 @@
...
+import marisa_trie
+import ujson
+
+from ..data_utils.almond_utils import quoted_pattern_with_space
+from .base import AbstractEntityDisambiguator
+
+logger = logging.getLogger(__name__)
+
+
+class NaiveEntityDisambiguator(AbstractEntityDisambiguator):
```
gcampax: How can the classes in this file not implement process_examples?

genienlp/ned/main.py
```diff
@@ -0,0 +1,190 @@
...
+    def __init__(self, args):
+        super().__init__(args)
+
+        with open(os.path.join(self.args.database_dir, 'es_material/alias2type.json'), 'r') as fin:
+            # alias2type.json is a big file (>4G); load it only when necessary
```
gcampax, line 213: Why is alias2type a JSON file at all, and not a binary-serialized marisa trie?

gcampax (changes requested):
This is definitely a good cleanup, and we're getting there, but we can still do better.

genienlp/ned/base.py
Mehrad0711: Ok yeah I should create a base class for Naive and Oracle and many of these methods should move there.

genienlp/ned/base.py
Mehrad0711: moved to base class

genienlp/ned/main.py
Mehrad0711: will get the method from abstract

genienlp/ned/ned_utils.py
Mehrad0711: yes. that's why there's the option of passing `ned_class_name` as an argument to override what args say.
Alternatively, we can pass a `mode` which can be (server|batch) but I avoided that since server mode and batch mode for non bootleg classes is the same.

genienlp/ned/main.py
Mehrad0711: Good point. I will convert it.

genienlp/ned/bootleg.py
Mehrad0711: Checkpoint name used to be different for different bootleg models. Now it's all the same. We still can pass different bootleg_models.

genienlp/ned/bootleg.py
Mehrad0711: We discussed this on slack. `ServingBootlegEntityDisambiguator` uses several methods and attributes from `BatchBootlegEntityDisambiguator`. Turning them into utility functions is not viable since methods are called within each other.

genienlp/ned/bootleg.py
Mehrad0711: Oracle uses `wiki2normalized_type` to normalize types too. Except only uses the first half of this function.  I have defined a base `normalize_types` method that other classes would inherit in a new PR I'm working on. That would solve this confusion.

gcampax (approved):
Alright, I think this is a very nice cleanup, let's get it in!

PR #189, opened 2021-08-09
Fix percentage problems in KF metrics

kevintangzero:
Unify the units of numbers in KF metrics.
See also: https://github.com/stanford-oval/thingpedia-common-devices/blob/e98aaaead21baced8115c34856a1a18a8bb84030/scripts/write_ui_metrics_outputs.py#L38

### DIFF
--- a/genienlp/write_kf_metrics.py
+++ b/genienlp/write_kf_metrics.py
@@ -60,7 +60,7 @@ def main(args):
         extracted_metrics = [
             {
                 'name': split + '-' + key,
-                'numberValue': val / 100,
+                'numberValue': val,
                 'format': "PERCENTAGE",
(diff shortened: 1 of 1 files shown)


gcampax:
Be careful with the comparison to the other file: this one uses metrics computed by `genienlp predict`, that one by `genie evaluate-server`. They are different metrics. I suspect `genienlp predict` uses 0-100 range for the metrics, not 0-1. See for example https://github.com/stanford-oval/genienlp/blob/master/genienlp/metrics.py#L273, which multiplies by 100

kevintangzero:
Yeah, I think we should somehow unify these numbers. But in this case, I'll close this pull request first.

Issue #174, opened 2021-07-14
Crash in Bootleg during request


gcampax:
The following command: "find me a movie with chris pratt" seems to reliably trigger a crash in Booleg with the currently deployed model on staging.

```
[E 210714 17:56:44 web:1789] Uncaught exception POST /v1/models/x40org-thingpedia-models-defaultx2fen:predict (127.0.0.1)
    HTTPServerRequest(protocol='http', host='x40org-thingpedia-models-defaultx2fen-predictor-default.staging.svc.cluster.local', method='POST', uri='/v1/models/x40org-thingpedia-models-defaultx2fen:predict', version='HTTP/1.1', remote_ip='127.0.0.1')
    Traceback (most recent call last):
      File "/usr/local/lib64/python3.8/site-packages/tornado/web.py", line 1704, in _execute
        result = await result
      File "/usr/local/lib/python3.8/site-packages/kfserving/handlers/http.py", line 79, in post
        response = (await model.predict(request)) if inspect.iscoroutinefunction(model.predict) else model.predict(request)
      File "/opt/genienlp/genienlp/kfserver.py", line 55, in predict
        results = self.server.handle_request(request)
      File "/opt/genienlp/genienlp/server.py", line 142, in handle_request
        output = generate_with_model(
      File "/opt/genienlp/genienlp/validate.py", line 60, in generate_with_model
        return generate_with_seq2seq_model(
      File "/opt/genienlp/genienlp/validate.py", line 124, in generate_with_seq2seq_model
        generated = model.generate(
      File "/opt/genienlp/genienlp/models/transformer_seq2seq.py", line 175, in generate
        generated = self.model.generate(
      File "/usr/local/lib64/python3.8/site-packages/torch/autograd/grad_mode.py", line 26, in decorate_context
        return func(*args, **kwargs)
      File "/usr/local/lib/python3.8/site-packages/transformers/generation_utils.py", line 970, in generate
        return self.greedy_search(
      File "/usr/local/lib/python3.8/site-packages/transformers/generation_utils.py", line 1327, in greedy_search
        if unfinished_sequences.max() == 0 or stopping_criteria(input_ids, scores):
    RuntimeError: CUDA error: device-side assert triggered
[E 210714 17:56:44 web:2239] 500 POST /v1/models/x40org-thingpedia-models-defaultx2fen:predict (127.0.0.1) 395.41ms
[...]

PR #190, opened 2021-08-09
Catch cuda runtime errors

kevintangzero:


genienlp/server.py
```diff
@@ -162,46 +162,49 @@ def handle_request(self, request):
...
+                    if sum(self.args.num_outputs) > 1:
+                        response = []
+                        for idx, predictions in enumerate(output.predictions):
+                            candidates = []
+                            for cand in predictions:
+                                candidates.append({'answer': cand})
+                            response.append({'candidates': candidates})
+                    else:
+                        response = [{'answer': p[0]} for p in output.predictions]
+        except RuntimeError:
```
gcampax: How do we check if this is a CUDA error or not?

kevintangzero: I added a test script, and the exit code 100 was caught.

s-jse: I think the problem is that other non-CUDA errors can cause the same exit code. Fairseq does something odd to differentiate CUDA errors from the rest (https://github.com/pytorch/fairseq/blob/50a671f78d0c8de0392f924180db72ac9b41b801/fairseq/trainer.py#L283):

```
except RuntimeError as e:
  if 'out of memory' in str(e):
    # do something about CUDA out of memory error
  else:
    raise e
```

gcampax: Yeah, except we care mostly about "device-side asserts" (ie page faults) rather than CUDA OOM. I'm not sure what to look for in the error message.

gcampax: #174 has some examples of stack traces that we care about

s-jse: Yeah, so checking `if 'RuntimeError: CUDA error: device-side assert triggered' in str(e)` should do the trick, although not in a very clean way.

kevintangzero: Nice to see the RuntimeError log. I found that the long sequence in my test script actually produced another RuntimeError: `RuntimeError: CUDA error: CUBLAS_STATUS_NOT_INITIALIZED when calling cublasCreate(handle)`, which is different from `device-side assert triggered`.

gcampax: We might want to catch "CUDA error" generically then.

kevintangzero: Addressed in the new update.

kevintangzero: I disabled the cuda tests in travis because travis is a cpu server.

gcampax (approved):
Yeah let's merge this.

s-jse (approved):

PR #191, opened 2021-08-16
Misc. updates for NED

Mehrad0711:


genienlp/arguments.py
```diff
@@ -427,7 +423,12 @@ def parse_argv(parser):
         default=0.3,
         help='Probability threshold for accepting a candidate for a mention',
     )
-    parser.add_argument('--bootleg_post_process_types', action='store_true', help='Postprocess bootleg types')
+    parser.add_argument(
+        '--ned_normalize_types',
+        type=str,
+        choices=['no', 'yes', 'force'],
```
s-jse: `--ned_type_normalization` with `choices=['off', 'lenient', 'strict']` makes it more obvious what this argument does.

genienlp/ned/ned_utils.py
```diff
@@ -230,6 +231,9 @@
         'publication date',
         'home address',
         'home assistant',
+        'key words',
+        'keywords',
+        'list of books',
```
s-jse: Probably cleaner if this list of banned words is kept in a separate json file.

genienlp/arguments.py
```diff
@@ -401,12 +401,8 @@ def parse_argv(parser):
...
     # NED args
     parser.add_argument('--do_ned', action='store_true', help='Collect and use entity features during training')
-    parser.add_argument(
-        '--min_entity_len', type=int, default=1, help='Minimum length for entities when ngrams database_lookup_method is used '
-    )
-    parser.add_argument(
-        '--max_entity_len', type=int, default=4, help='Maximum length for entities when ngrams database_lookup_method is used '
-    )
+    parser.add_argument('--min_entity_len', type=int, default=1, help='Minimum token-length of entities in ngram-based lookup')
+    parser.add_argument('--max_entity_len', type=int, default=4, help='Maximum token-length of entities in ngram-based lookup')
```
gcampax: I am like 50% confident that I saw this argument in Bootleg code as well, so why this help text talks about ngram-based lookup?

genienlp/arguments.py
```diff
@@ -427,7 +423,12 @@ def parse_argv(parser):
         default=0.3,
         help='Probability threshold for accepting a candidate for a mention',
     )
-    parser.add_argument('--bootleg_post_process_types', action='store_true', help='Postprocess bootleg types')
+    parser.add_argument(
+        '--ned_normalize_types',
+        type=str,
+        choices=['no', 'yes', 'force'],
+        help='Normalize types. yes: attempt to map; if unsuccessful use original. force: attempt to map; if unsuccessful drop the type.',
+    )
```
gcampax, line 441: What's the equivalent of the previous option here? The config backcompat code says "yes", but I thought the previous logic was "force"

genienlp/ned/bootleg.py
```diff
@@ -64,10 +63,13 @@ def __init__(self, args):
 
         with open(f'{self.args.database_dir}/wiki_entity_data/type_mappings/wiki/qid2typenames.json') as fin:
             self.entityqid2typenames = ujson.load(fin)
+
+        # large print
+        self.entityqid2typenames['Q1548123'] = ['format']
```
gcampax: Uh? Why is this here? In the code?

genienlp/ned/bootleg.py
```diff
@@ -135,116 +137,110 @@ def extract_mentions(self, input_path):
...
+        # filter qids with probability lower than a threshold
+        idx = reverse_bisect_left(all_probs, self.args.bootleg_prob_threshold)
+        all_qids = all_qids[:idx]
+        all_probs = all_probs[:idx]
+
+        if len(all_qids) > self.args.max_qids_per_entity:
+            all_qids = all_qids[: self.args.max_qids_per_entity]
+            all_probs = all_probs[: self.args.max_qids_per_entity]
+
+        if not is_banned(alias):
```
gcampax:
```
if is_banned(alias):
    return
```

and you save one indentation level everywhere, making things more readable.

genienlp/ned/scripts/analyze_bootleg_results.py
```diff
@@ -0,0 +1,97 @@
...
+    all_qids = []
+    all_probs = []
+    unknown_qids = set()
+
+    lines = jsonlines.open(args.input_file, 'r')
+
+    all_titles = Counter()
+    all_new_titles = Counter()
+
+    fout = open('results.txt', 'w')
```
gcampax: This should be a command-line argument. You cannot assume you can write to the current directory.

genienlp/ned/scripts/analyze_bootleg_results.py
```diff
@@ -0,0 +1,97 @@
...
+
+parser.add_argument('--input_file', type=str)
+parser.add_argument('--database_dir', type=str)
+parser.add_argument('--ned_domains', type=str, nargs='+')
+parser.add_argument('--subsample', type=int, default='1000000000')
+
+
+args = parser.parse_args()
+
+if __name__ == '__main__':
```
gcampax: Why is this not a normal genienlp command?

genienlp/run_bootleg.py
```diff
@@ -96,10 +96,10 @@ def parse_argv(parser):
     )
 
     parser.add_argument(
-        '--min_entity_len', type=int, default=1, help='Minimum length for entities when ngrams database_lookup_method is used '
+        '--min_entity_len', type=int, default=1, help='Minimum token-length of entities in ngram-based lookup of bootleg'
```
gcampax: "ngram-based lookup" is implementation detail, the point is that entities shorter than that will not be retrieved

genienlp/ned/bootleg.py
```diff
@@ -135,116 +137,110 @@ def extract_mentions(self, input_path):
...
+                # to map qids to unique ids we just need to remove the Q character as qids are distinct
+                qids.append(int(qid[1:]))
+
+                # get all types for a qid
+                all_types = []
+                if 'books' in self.args.ned_domains:
+                    if alias in ['houghton mifflin', 'ciudad de buenos aires']:
+                        all_types = ['award']
+                    elif alias in ['blanche', 'biblioteca breve', '3rd edition', '4th edition']:
+                        all_types = ['version, edition, or translation']
```
gcampax: This is not the place for this stuff.

Also, to have this stuff at all in Bootleg is meaningless: you're just raising your accuracy on the dev set by looking at oracle results and cleaning up by hand, it's not scientific.

tests/test_NED.sh
```diff
@@ -22,7 +22,7 @@ do
 
     # check if result file exists
     if test ! -f $workdir/model_$i/eval_results/valid/almond_dialogue_nlu.tsv ; then
-        echo "File not found!"
+        echo "File not found!"database_lookup_method
```
gcampax: Mistake here?

gcampax (changes requested):
Btw, as a general rule, you should rebase/squash patches like "make X work again" or "fixed to make tests pass" into whatever patch broke things. Ideally, tests should pass after _every_ commit, so if there are regressions not caught by the tests we can bisect and test all commits to find the culprit.

genienlp/ned/scripts/oracle_vs_bootleg.py
```diff
@@ -0,0 +1,98 @@
+import argparse
```
s-jse: Should this and `analyze_bootleg_results` be added as genienlp scripts?

s-jse (approved):

genienlp/arguments.py
Mehrad0711: Yes. This argument is shared between the two scripts. Here it's used for the naive NED approach which does NER right before training (as opposed to bootleg which does it in a separate step by running run_bootleg.py)

genienlp/arguments.py
Mehrad0711: previous approach was using "yes" logic.

genienlp/ned/bootleg.py
Mehrad0711: `qid2typenames` map qids to an unordered list of types. I noticed the type we want is sometimes not ranked highly. So I have to do some reordering. I don't want to modify the files directly cause I'd forget what I changed later. Instead do it in the code for visibility and version tracking.

genienlp/ned/scripts/analyze_bootleg_results.py
Mehrad0711: These scripts are for local data analysis. I don't want to crowd genienlp commands with data processing/ analysis scripts.

genienlp/ned/bootleg.py
Mehrad0711: I'm reaching the conclusion that for certain entities we need to map them directly to the correct type.
It happens in 2 cases:
1) bootleg partly detected the entity and thus got the Qid wrong. 
2) bootleg alias2qids has incorrect entries where the book’s writer Qid is associated with the book name instead of the book itself, thus mapping to “person” type instead of “book”.

In none of those cases changing type mapping will fix the problem. Also, this mapping is only done for training set, not dev/ test set.
I added this as if-else in the code to see how experiments turn out. I'm eventually gonna move it to a json file similar to type mapping.

genienlp/ned/scripts/oracle_vs_bootleg.py
Mehrad0711: These scripts are for local data analysis. I don't want to crowd genienlp commands with data processing/ analysis scripts.

Mehrad0711:
> Btw, as a general rule, you should rebase/squash patches like "make X work again" or "fixed to make tests pass" into whatever patch broke things. Ideally, tests should pass after _every_ commit, so if there are regressions not caught by the tests we can bisect and test all commits to find the culprit.

Believe me, I tried hard to make these PRs presentable and keep a clean commit history! In some cases, there were multiple fixes in one commit and it was hard (not impossible) to break it and merge with commits that introduced a bug. 
I agree we should strive for buildable commits though esp. now that more people will be contributing to the code.

genienlp/arguments.py
gcampax: This is confusing. How about "Minimum token-length of entities (does not apply to Bootleg)"?

genienlp/ned/bootleg.py
gcampax: This is the wrong way around it. Changing the code is harder. We don't want hacks in the code that confuse readers and contributors. If you want to version the bootleg input files, you can do it in the workdirs (with some light-weight override file, not by tracking the full bootleg file)

genienlp/ned/scripts/analyze_bootleg_results.py
gcampax: This is a generally usable script that could be part of everyone's workflow. It's not a custom one-off script. I think it's worth making it part of genienlp proper.

genienlp/ned/bootleg.py
gcampax: Yeah don't merge this hacks to the code. Add an override file if you need to, but I still don't think you need to.

genienlp/ned/scripts/analyze_bootleg_results.py
Mehrad0711: My main concern is this will be an invitation for more scripts. As a general rule, I think we should exclude data manipulation scripts that do not rely on a specific model/ tokenizer and have them in genie-toolkit or workdirs. I can add these for now since you and Sina are both in favor and Kevin may wanna take a look too.

gcampax: I agree, anything that's project/model specific should go into genie-workdirs.

gcampax (approved):
Ok to merge when tests are happy

PR #196, opened 2021-08-17
Merge nrser/first-deploy branch -- fix Docker builds

nrser:
Main change is bumping `Mehrad0711/bootleg` dep to a new commit that relaxes the pydantic requirement -- `pip`'s new resolver was going pathological over the old spec.

### DIFF
--- a/dev/bin/docker.build.sh
+++ b/dev/bin/docker.build.sh
@@ -1,0 +1,16 @@
+#!/usr/bin/env bash
+
+# Common / useful `set` commands
+set -Ee # Exit on error
+set -o pipefail # Check status of piped commands
+set -u # Error on undefined vars
+# set -v # Print everything
+# set -x # Print commands (with expanded vars)
+
+cd "$(git rev-parse --show-toplevel)"
+
+docker build \
+	--build-arg BASE_IMAGE=nvidia/cuda:10.1-runtime-ubi8 \
+	-f ./dockerfiles/Dockerfile \
+	-t "stanfordoval/genienlp:$(git describe)-cuda" \
+	.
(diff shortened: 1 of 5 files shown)


dev/bin/docker.build.sh
```diff
@@ -0,0 +1,16 @@
...
+# Common / useful `set` commands
+set -Ee # Exit on error
+set -o pipefail # Check status of piped commands
+set -u # Error on undefined vars
+# set -v # Print everything
+# set -x # Print commands (with expanded vars)
+
+cd "$(git rev-parse --show-toplevel)"
+
+docker build \
```
Mehrad0711, line 12: Is this file doing sth that the [docker build](https://github.com/stanford-oval/genienlp/blob/master/dockerfiles/hooks/build) script does not? Perhaps it could be added as another case statement?

dev/bin/pyenv.reset.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
```
Mehrad0711, line 1: Is this file for local use or needed in the docker image? In the former case perhaps it shouldn't be committed to genienlp.

setup.py
```diff
@@ -65,7 +65,7 @@
         # for kf
         'kfserving>=0.5.0',
         # for NED
-        'bootleg @ git+https://github.com/Mehrad0711/bootleg@b6c207259a83c21c77aeaf8088ee85f24ec7e708',
+        'bootleg @ git+https://github.com/Mehrad0711/bootleg@08da02ddc10540688747291f8ab54a1bacdfe433',
```
Mehrad0711, line 68: I bumped bootleg to 1.0.5 which includes the newer version for pydantic so no changes will be needed here.

Mehrad0711:
Closing this...
Feel free to open if there are further updates.

PR #204, opened 2021-09-27
Wip/mehrad/updates

Mehrad0711:
The bitod branch is accumulating different changes and growing. Before it becomes too late, I'm gonna break them down and emit each big change one at a time.

Changes in this PR:
- Compute and return validation loss (can be used as decascore to choose the best checkpoint based on lowest loss instead of highest "accuracy")
- Some tweaks to reduce GPU memory usage so we can fit bigger batch sizes for training/ inference
- Misc. code updates

### DIFF
--- a/genienlp/models/base.py
+++ b/genienlp/models/base.py
@@ -33,6 +33,8 @@ import os
 import torch
 from transformers import PreTrainedModel
 
+from genienlp.tasks.almond_task import Translate
+
 from ..data_utils.numericalizer import TransformerNumericalizer
 
 logger = logging.getLogger(__name__)
@@ -77,3 +79,8 @@ class GenieModel(PreTrainedModel):
         self.numericalizer.grow_vocab(tasks)
         if self.numericalizer.num_tokens > old_num_tokens:
             logger.info(f'Vocabulary has expanded to {self.numericalizer.num_tokens} tokens')
+
+    def set_task_dependent_generation_kwargs(self, tasks):
+        self._output_scores = any('loss' in task.metrics for task in tasks)
+        self._output_attentions = any(isinstance(task, Translate) for task in tasks)
+        self._output_hidden_states = False
--- a/genienlp/models/base.py
+++ b/genienlp/models/base.py
@@ -33,8 +33,6 @@ import os
 import torch
 from transformers import PreTrainedModel
 
-from genienlp.tasks.almond_task import Translate
-
 from ..data_utils.numericalizer import TransformerNumericalizer
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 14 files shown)


genienlp/models/base.py
```diff
@@ -77,3 +79,8 @@ def add_new_vocab_from_data(self, tasks, resize_decoder=False):
         self.numericalizer.grow_vocab(tasks)
         if self.numericalizer.num_tokens > old_num_tokens:
             logger.info(f'Vocabulary has expanded to {self.numericalizer.num_tokens} tokens')
+
+    def set_task_dependent_generation_kwargs(self, tasks):
+        self._output_scores = any('loss' in task.metrics for task in tasks)
+        self._output_attentions = any(isinstance(task, Translate) for task in tasks)
+        self._output_hidden_states = False
```
gcampax, line 84: If this is task dependent, then surely it should live on the Task, and not on the model?
More generally, model code should be independent on the task.

Also, should we just unconditionally set these to true (except _output_scores maybe)? Is there a performance loss if we do that?

genienlp/predict.py
```diff
@@ -38,6 +38,8 @@
 
 from torch.multiprocessing import Process, set_start_method
 
+from genienlp.metrics import calculate_and_reduce_metrics
```
gcampax: Use relative imports.

genienlp/train.py
```diff
@@ -242,8 +235,8 @@ def do_validate(
         output, metric_dict = validate(val_task, val_iter, model, numericalizer, args, num_print=args.num_print)
         val_loss = output.loss
         if val_loss is not None:
-            log_entry = f'{args.timestamp}:{elapsed_time(logger)}:iteration_{iteration}:{round_progress}train_{train_task.name}:{task_progress}val_{val_task.name}:val_loss{val_loss.item():.4f}:'
-            writer.add_scalar(f'loss/{val_task.name}/val', val_loss.item(), iteration)
+            log_entry = f'{args.timestamp}:{elapsed_time(logger)}:iteration_{iteration}:{round_progress}train_{train_task.name}:{task_progress}val_{val_task.name}:val_loss_{val_loss:.4f}:'
+            writer.add_scalar(f'loss/{val_task.name}/val', val_loss, iteration)
```
gcampax, line 239: val_loss is a Tensor, don't you need this change?

genienlp/validate.py
```diff
@@ -136,11 +140,12 @@ def generate_with_seq2seq_model(
                 do_sample=args.temperature[hyperparameter_idx] != 0,  # if temperature==0, we do not sample
             )
             partial_batch_prediction_ids = generated.sequences
-            cross_attentions = getattr(generated, 'cross_attentions', None)
 
-            if cross_attentions is not None:
+            if model._output_attentions:
```
gcampax: _ attributes are private, and cannot be accessed outside of a class.

genienlp/train.py
Mehrad0711: val_loss is already turned to scaler inside validate.py

genienlp/models/base.py
Mehrad0711: > If this is task dependent, then surely it should live on the Task, and not on the model?

We're already passing tasks to model init for numericalizer to read the special tokens. Also not sure why it should be in Task. Based on task, we need to adjust some attributes of the model. We still need to pass some information from task to model if not the whole object.

> Also, should we just unconditionally set these to true (except _output_scores maybe)? Is there a performance loss if we do that?

It makes a difference in memory usage. With all of them being False, we can fit bigger batches for inference.

gcampax: You don't want to depend on the specific Task instance, because next time someone will come along with another Translate-like class and it won't work for obscure reasons.
If you want, make it an attribute of the base Task class like `needs_attention_scores` or similar.

gcampax (approved):
Ok

PR #211, opened 2021-10-05
Wip/translation api

Mehrad0711:
Accommodating https://github.com/stanford-oval/genie-toolkit/pull/806

### DIFF
--- a/genienlp/server.py
+++ b/genienlp/server.py
@@ -50,6 +50,25 @@ from .validate import generate_with_model
 
 logger = logging.getLogger(__name__)
 
+GENERATION_ARGUMENTS = {
+    'num_beams',
+    'num_beam_groups',
+    'diversity_penalty',
+    'num_outputs',
+    'no_repeat_ngram_size',
+    'top_p',
+    'top_k',
+    'repetition_penalty',
+    'temperature',
+    'max_output_length',
+    'src_locale',
+    'tgt_locale',
+    'do_alignment',
+    'align_preserve_input_quotation',
+    'align_remove_output_quotation',
+    'translate_example_split',
+}
+
 
 def parse_argv(parser):
     parser.add_argument('--path', type=str, required=True)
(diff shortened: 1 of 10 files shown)


genienlp/server.py
```diff
@@ -120,8 +121,22 @@ def numericalize_examples(self, ex):
...
+            setattr(args, k, v)
+
+        # TODO handle this better by decoupling numericalizer and model
+        if hasattr(args, 'src_locale') and hasattr(args, 'tgt_locale'):
+            src_locale, tgt_locale = adjust_language_code(
+                self.model.config, self.args.pretrained_model, args.src_locale, args.tgt_locale
+            )
+            self.numericalizer._tokenizer.src_lang = src_locale
+            self.numericalizer._tokenizer.tgt_lang = tgt_locale
+            self.model.update_language_dependent_configs(tgt_locale)
```
gcampax, line 157: You should avoid accessing _tokenizer from outside the numericalizer class, it's a private property. Add new methods to Numericalizer instead.

gcampax (approved):
This looks good I think.

setup.py
```diff
@@ -64,7 +64,7 @@
         'mosestokenizer~=1.1',
         'pathos==0.2.8',
         # for kf:
-        'kfserving>=0.5.0',
+        'kfserving==0.5.1',
```
gcampax: Please don't do this. kfserving depends on ray, but doesn't specify a version bound on it. ray has broken APIs before, breaking kfserving, and us transitively. If we pin kfserving, we remain broken even if kfserving releases a fixed version, until we change genienlp. This creates a cascade effect and takes a lot of work to fix.
(yes python packaging is the worst)

Mehrad0711: what is the solution then? pinning "ray" instead?

gcampax: There is no good solution, but at least if we leave kfserving unpinned we'll get any fix they make as soon as they make it.

Mehrad0711: but that will leave our tests broken till then. Apparently ray 1.7.0 was released 3 days ago and had major changes. I suggest pinning ray to 1.6.0 until they patch 1.7.0. what do you think?

Mehrad0711:
Ok @gcampax, I pinned ray for now until they patch it. I'm gonna merge this branch and rebase the other one.

PR #216, opened 2021-10-21
Save disk space when saving models w/o finetuning

Mehrad0711:
Don't save optimizer state and replicated checkpoint


gcampax:
Is this for train_iterations=0 ?

Mehrad0711:
Correct

gcampax (approved):
Yeah

PR #221, opened 2021-10-29
train: don't map save_dict to cpu before saving

Mehrad0711:
PR pitch:

Have you ever noticed genienlp checkpoints suspiciously take twice the disk space of their corresponding pretrained model?
Well, if not, don't worry cause you won't notice it anymore!

### DIFF
--- a/genienlp/train.py
+++ b/genienlp/train.py
@@ -295,8 +295,6 @@ def maybe_save(
         # to load this model later
         model_state_dict = model.module.state_dict()
 
-    model_state_dict = {k: v.cpu() for k, v in model_state_dict.items()}
-
     save_model_state_dict = {'model_state_dict': model_state_dict, 'best_decascore': best_decascore}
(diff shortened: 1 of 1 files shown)


s-jse (approved):

gcampax:
Hold on, what's going on here? Why are CPU tensors more expensive than GPU tensors? And what happens if we try to load on CPU a GPU tensor? I thought PyTorch would just blow up in our face.

Mehrad0711:
Loading is not affected. We're still loading models on the available device using `map_location` regardless of tensor devices in save_dict.
Moving tensors to .cpu() before saving is coming from decaNLP. It was probably necessary for the older torch versions but not anymore (check implementations in huggingface). What is very very weird though is checkpoints saved on CPU vs GPU before changes in this PR have the exact same content except one is double the size! My guess is there is meta info attached to the saved files (not user-visible) causing this which is likely a bug in torch.

Issue #228, opened 2021-11-24
Prediction with TransformerLSTM fails


gcampax:
Caught by chance in the Travis logs of genie-toolkit, which trains a TransformerLSTM for schema2qa and predicts with genienlp server:

```
Traceback (most recent call last):
  File "/home/travis/.local/bin/genienlp", line 8, in <module>
    sys.exit(main())
  File "/home/travis/.local/lib/python3.7/site-packages/genienlp/__main__.py", line 108, in main
    subcommands[argv.subcommand][2](argv)
  File "/home/travis/.local/lib/python3.7/site-packages/genienlp/server.py", line 382, in main
    server.run()
  File "/home/travis/.local/lib/python3.7/site-packages/genienlp/server.py", line 317, in run
    self._run_stdin()
  File "/home/travis/.local/lib/python3.7/site-packages/genienlp/server.py", line 306, in _run_stdin
    sys.stdout.write(self.handle_json_request(line))
  File "/home/travis/.local/lib/python3.7/site-packages/genienlp/server.py", line 267, in handle_json_request
    return json.dumps({'id': request['id'], 'instances': self.handle_request(request)}) + '\n'
  File "/home/travis/.local/lib/python3.7/site-packages/genienlp/server.py", line 252, in handle_request
    task, args = self._init_request(request)
  File "/home/travis/.local/lib/python3.7/site-packages/genienlp/server.py", line 157, in _init_request
    self.model.update_language_dependent_configs(tgt_locale)
  File "/home/travis/.local/lib/python3.7/site-packages/torch/nn/modules/module.py", line 1131, in __getattr__
    type(self).__name__, name))
AttributeError: 'TransformerLSTM' object has no attribute 'update_language_dependent_configs'
```

PR #229, opened 2021-11-24
Fix bug in server for TransformerLSTM

Mehrad0711:
Addressing #228

### DIFF
--- a/genienlp/server.py
+++ b/genienlp/server.py
@@ -231,9 +231,7 @@ class Server(object):
                         instance['score'][self.estimator_filenames[e_idx]] = float(estimator_scores[idx])
                     response.append(instance)
         else:
-            output = generate_with_model(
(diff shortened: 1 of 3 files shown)


genienlp/server.py
```diff
@@ -154,7 +154,8 @@ def _init_request(self, request):
                 self.model.config, self.args.pretrained_model, args.src_locale, args.tgt_locale
             )
             self.numericalizer.update_language_dependent_properties(src_locale, tgt_locale)
-            self.model.update_language_dependent_configs(tgt_locale)
+            if hasattr(self.model, 'update_language_dependent_configs'):
```
gcampax: Rather than this, it would be cleaner to add an empty implementation of the method in the base class.

gcampax (approved):
Cool

PR #234, opened 2021-12-15
Wip/misc fixes

Mehrad0711:
- Fixes for using validation "loss" as deca_score
- Allow setting "min_output_len" via cmdline. Models like MT5 need min_output_length of 2.
- Add polynomial lr scheduler which yields better results for MT5 than linear/ transformers schedulers.
- Raise exception when user-provided min/ max example lengths do not agree with actual data lengths (after tokenization)


genienlp/arguments.py
```diff
@@ -186,6 +186,13 @@ def parse_argv(parser):
     parser.add_argument('--eval_set_name', type=str, help='Evaluation dataset name to use during training')
 
     parser.add_argument('--max_output_length', default=150, type=int, help='maximum output length for generation')
+    parser.add_argument(
+        '--min_output_length',
+        default=3,
+        type=int,
+        help='maximum output length for generation; '
+        'default is 3 for most multilingual models: BOS, language code, and one token. otherwise it is 2',
```
s-jse, line 194: I think you are completely ignoring `[EOS]`. I think `min_output_length` can be set to 3 for everything. Originally, I set it to 3 for `BOS, one token, EOS` (no language code). Would the classification task in your experiments be fixed if you added a [EOS] token to the end of your training example like `[BOS] false [EOS]`?

genienlp/predict.py
```diff
@@ -164,7 +164,13 @@ def parse_argv(parser):
         default=[0],
         help='ngrams of this size cannot be repeated in the output. 0 disables it.',
     )
-    parser.add_argument('--max_output_length', default=150, type=int, help='maximum output length for generation')
+    parser.add_argument('--max_output_length', type=int, help='maximum output length for generation')
```
s-jse, line 167: Removing the default value for `max_output_length` can -under rare circumstances- result in some paraphrasing models requiring a smaller prediction batch size. So please revert this change.

s-jse (commented):
Good overall. I left a few comments.

genienlp/arguments.py
gcampax: Hold on, why is BOS counted in the generation length? We don't generate BOS, it goes into the input.

Mehrad0711: @gcampax  Have you forgotten what you wrote [here](https://github.com/stanford-oval/genienlp/blob/e4f9c80fc21030dad0137b05fd8f40d6d3830456/genienlp/models/transformer_seq2seq.py#L128) ;p 
All pretrained BART checkpoints generate `[BOS] X [EOS]`. For bart-large I think you did some experiments and decided it works better if we finetune without [BOS] in the output. It's generally fine since we finetune and model learns not to generate [BOS].
MBART generates `X [EOS] [lang_code]`. MT5/ T5 generates `X [EOS]`. 
Before introducing multilingual models, min_output_length was hard-coded to 2. After becoming multilingual I made it 3. 
Now trying T5 realized we should do 2 but for backward compatibility with previously trained models, I think we should keep the default 3 and pass what we want for other models.
If we pass the wrong one, it will be caught later in the length exceptions I added in [this commit ](https://github.com/stanford-oval/genienlp/pull/234/commits/c709f72120ed7a890a379645382942de0f07d020)

genienlp/predict.py
Mehrad0711: min/max_output_length should match with what was provided for training right? (If we provide defaults here they will override train args.)

Mehrad0711:
Ok is there anything else here? Can I merge it?

PR #235, opened 2021-12-16
Redo bootleg analysis scripts

gcampax:
This PR reimplements the type mapping for bootleg, and redoes the analysis scripts.
I haven't touched the existing type mapping files. The PR depends on https://github.com/stanford-oval/thingpedia-common-devices/pull/458 as thingpedia-common-devices will be the new home of type mapping files. Most importantly, you need two separate type mapping files, one for bootleg and one for oracle.

Results are at https://docs.google.com/spreadsheets/d/159oZV2aE4Jy7lTzyweSIuy03tU4rh1HPPIr8M8A63wo/edit#gid=33582497
We get about 58% F1 on the training set and 76% F1 on the dev set.
Precision is in general very low, and particularly low on the training set, because there are many confounding entities: entities that are known to bootleg but not to the oracle (for example they appear in a search query).

### DIFF
--- a/genienlp/train.py
+++ b/genienlp/train.py
@@ -659,6 +659,13 @@ def main(args):
     src_lang = args.train_src_languages.split('+')[0]
     tgt_lang = args.train_tgt_languages.split('+')[0]
 
+    # dump entities if required
+    if args.ned_dump_entity_type_pairs and args.add_entities_to_text == 'append':
+        for task, train_set, val_set in zip(tasks, train_sets, val_sets):
+            ned_dump_entity_type_pairs(train_set, args.data, 'train', task.utterance_field)
+            ned_dump_entity_type_pairs(val_set, args.data, 'eval', task.utterance_field)
+        return
+
     ########## initialize model
     best_decascore = None
     if args.load is not None:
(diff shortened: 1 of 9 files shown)


genienlp/train.py
```diff
@@ -658,6 +658,13 @@ def main(args):
     src_lang = args.train_src_languages.split('+')[0]
     tgt_lang = args.train_tgt_languages.split('+')[0]
 
+    # dump entities if required
+    if args.ned_dump_entity_type_pairs and args.add_entities_to_text == 'append':
+        for task, train_set, val_set in zip(tasks, train_sets, val_sets):
+            ned_dump_entity_type_pairs(train_set, args.data, 'train', task.utterance_field)
+            ned_dump_entity_type_pairs(val_set, args.data, 'eval', task.utterance_field)
+        return
```
Mehrad0711: I would remove the return here since exiting code in training with no warnings can cause confusions in the future when debugging. Also, dumping can be done before training an actual model when running oracle - we don't wanna run the code twice.

Mehrad0711:
@gcampax this looks good. Should I take the liberty of addressing my comment, polishing, and merging this PR?

PR #250, opened 2022-02-19
Fix prediction with `num_beams > 1` and `num_outputs > 1`

s-jse:


genienlp/validate.py
```diff
@@ -410,7 +410,7 @@ def validate(task, val_iter, model, numericalizer, args, num_print=10):
             output.predictions, output.answers, metrics_to_return, args.reduce_metrics, model.tgt_lang
         )
 
-        results = {'beam search': output.predictions, 'answer': output.answers, 'context': output.contexts}
+        results = {'model prediction': output.predictions, 'answer': output.answers, 'context': output.contexts}
```
Mehrad0711: Can we use "prediction" instead to be consistent with other keys?

genienlp/predict.py
```diff
@@ -461,17 +466,26 @@ def run(args, device):
...
+                            + generation_output.answers[i]
+                        )]  # one line with all generation outputs separated by '\t'
                     if args.calibrator_paths is not None:
                         for score in generation_output.confidence_scores:
-                            line += '\t' + str(score[i])
-                    prediction_file.write(line + '\n')
+                            lines = [line + '\t' + str(score[i]) for line in lines] # append score to all lines
+                    prediction_file.write('\n'.join(lines) + '\n')
 
             if args.translate_return_raw_outputs:
```
Mehrad0711, line 492: Can we have "one_output_per_line" option for this if branch too?

genienlp/validate.py
Mehrad0711: Also to make printed results align nicely in terminal, we use the maximum number of characters among all keys: https://github.com/stanford-oval/genienlp/blob/1d3f64bfa4d5077c5fe2748977d88ccdc16dd548/genienlp/validate.py#L392 
Maybe it's time to change the "11" to `max(len(key) for key in results)`

s-jse: I think "prediction" and "answer" are vague to first time users, as answer in some other code bases means the answer the model gives. So I changed them to "model prediction" and "gold answer".

Mehrad0711: yeah, this is another one of those things we need to standardize throughout the code. We have "beam search" here and "greedy" in metrics.py!
looks good for this PR, but should we go with "pred" and "gold" everywhere?

genienlp/predict.py
```diff
@@ -490,14 +490,23 @@ def run(args, device):
...
                     for i in range(len(generation_output.example_ids)):
-                        line = (
-                            generation_output.example_ids[i]
-                            + '\t'
-                            + '\t'.join(generation_output.raw_predictions[i])
-                            + '\t'
-                            + generation_output.answers[i]
-                        )  # all outputs separated by '\t'
-                        prediction_file.write(line + '\n')
+                        if args.one_output_per_line:
```
Mehrad0711, line 495: Thanks. I'll remove this branching thing too. We should add a function for creating the correct output.

Mehrad0711 (approved):

PR #251, opened 2022-02-25
Support end-to-end dialogue training and evaluation

Mehrad0711:
The code provides a data-dependent interface for training and evaluating dialogues end-to-end (e.g. bitod).
The [dialogues](https://github.com/Mehrad0711/dialogues) repository will host the data-specific code: data processing, API calling, independent evaluations.

@s-jse we discussed some of the changes. If there are any small or non-breaking changes I will address them in this PR. Otherwise, let's add them to [cleanup items](https://github.com/stanford-oval/genienlp/projects/4), and I will send a separate PR.


setup.py
```diff
@@ -68,5 +68,7 @@
         'scikit-learn>=0.23,<2.0',
         'dill~=0.3',
         'xgboost~=1.3',
+        # dialogues
+        'dialogues @ git+https://github.com/Mehrad0711/dialogues',
```
s-jse, line 72: We should make this optional since most people who use GenieNLP won't need this. Like `matplotlib`.

genienlp/paraphrase/scripts/transform_dataset.py
```diff
@@ -144,7 +144,7 @@ def main(args):
             new_queries = []  # list of lists
             query_file = open(args.query_file, 'r')
             for line in query_file:
-                queries = line.split('\t')[1:-1]  # 0 is example id, -1 is gold answer
+                queries = line.split('\t')[1:-2]  # 0 is example id, -1 is input, -2 is gold answer
```
s-jse, line 147: Why did you change this? Doesn't it break other paraphrasing scripts?

genienlp/data_utils/iterator.py
```diff
@@ -57,7 +57,7 @@ def __init__(self, data_source, batch_size, sort, shuffle_and_repeat, sort_key_f
         self.batch_size_fn = batch_size_fn
         self.groups = groups
 
-        if sort:
+        if sort and self.sort_key:
```
s-jse: Did it crash without this check? If so, there is a bug somewhere else as well.

genienlp/metrics.py
```diff
@@ -513,7 +514,74 @@ def computeDialogue(greedy, answer):
     return joint_goal_em, turn_request_em, turn_goal_em, answer
 
 
-def compute_metrics(greedy, answer, requested_metrics: Iterable, lang):
+def computeBITOD(greedy, answer, tgt_lang, args, example_ids):
```
s-jse: Rename `computeBITOD` to something more general like `compute_e2edialogue_score`.

genienlp/metrics.py
```diff
@@ -532,6 +600,15 @@ def compute_metrics(greedy, answer, requested_metrics: Iterable, lang):
     metric_values = []
     if not isinstance(answer[0], list):
         answer = [[a] for a in answer]
+    if 'e2e_dialogue_score' in requested_metrics:
+        requested_metrics += ['JGA', 'API_em', 'DA_em', 'BLEU']
+        results = computeBITOD(greedy, answer, lang, args, example_ids)
+        metric_keys += results.keys()
+        metric_values += results.values()
+    if 'jga' in requested_metrics:
+        JGA = computeJGA(greedy, answer, example_ids)
```
s-jse: Use lowercase for variable names like `JGA` to be consistent.

genienlp/metrics.py
```diff
@@ -513,7 +514,74 @@ def computeDialogue(greedy, answer):
...
+        convert_lists_to_set(answer_state)
+        convert_lists_to_set(greedy_state)
+
+        if answer_state == greedy_state:
+            hit += 1
+
+    return hit / len(greedy) * 100
+
+
+def compute_metrics(greedy, answer, requested_metrics, lang, args, example_ids=None):
```
s-jse: Rebase this branch on `master`. This function will have conflicts because I changed it in #250.

genienlp/predict.py
```diff
@@ -105,6 +106,12 @@ def parse_argv(parser):
     parser.add_argument('--cache', default='.cache', type=str, help='where to save cached files')
     parser.add_argument('--subsample', default=20000000, type=int, help='subsample the eval/test datasets')
 
+    parser.add_argument(
+        '--pred_file',
+        type=str,
+        help='If provided, we just compute evaluation metrics on it and bypass model prediction. File should be in tsv format with id, pred, target columns',
```
s-jse: id, prediction, **answer** not **target**!

s-jse: This should be a separate genienlp command in a separate python file, something like `genienlp calculate_metric --pred_file ...` and `calculate_metric.py`. Because in the current implementation you are practically bypassing almost all of predict.py and defining functions like `compute_metrics_on_file` that have nothing to do with normal prediction. It's better if you just move everything into a separate file.

genienlp/predict.py
```diff
@@ -255,6 +286,10 @@ def set_default_values(args):
     if args.confidence_feature_path is None:
         args.confidence_feature_path = os.path.join(args.path, 'confidence_features.pkl')
 
+    if args.e2e_dialogue_evaluation and args.val_batch_size[0] != 1:
+        logger.warning('When evaluating bitod end2end val_batch_size should be 1 so we load data turn by turn')
```
s-jse: **the** data?

genienlp/tasks/generic_task.py
```diff
@@ -393,3 +393,98 @@ def metrics(self):
...
+        super().__init__(name, args)
+        self._metrics = ['casedbleu']
+
+    def get_splits(self, root, **kwargs):
+        kwargs['train_target'] = 'rg'
+        kwargs['e2e_evaluation'] = self.args.e2e_dialogue_evaluation
+        return BiTODDataset.return_splits(path=root, make_example=self._make_example, **kwargs)
+
+
+@register_task('e2e_dialogue_dst')
```
s-jse: This and `BiTODNLG` tasks are confusing to me. Is this BiTOD specific or general? If it is general, you should rename them to something more general, but then you are calling `BiTODDataset` methods, which is definitely dataset specific.

genienlp/util.py
```diff
@@ -599,9 +599,16 @@ def make_data_loader(dataset, numericalizer, batch_size, device=None, train=Fals
         f'answer lengths (min, mean, max): {np.min(answer_lengths)}, {int(np.mean(answer_lengths))}, {np.max(answer_lengths)}'
     )
 
-    if dataset.batch_size_fn == input_tokens_fn:
+    if train:
```
s-jse, line 602: Can you explain the design here? Is the different sorting something you need for e2e dialogues?

s-jse (changes requested):
Overall looks good, please see the individual comments. Also, please rebase on `master` and add tests.

genienlp/util.py
Mehrad0711: train and eval sets may need different sorting or batching. for example in e2e dialogues, eval set should not be sorted and should be batched based on number of examples instead of tokens.

genienlp/data_utils/iterator.py
Mehrad0711: I moved this outside to the function calling. This is to check whether to sort the dataset or not.
If we don't want the dataset to be stored, we should set `sort_key_fn` or `eval_sort_key_fn` to "None" in the Dataset class.
e.g. https://github.com/stanford-oval/genienlp/blob/93de0965e22fe74257aadad561a7eae2a8bbf26e/genienlp/tasks/generic_dataset.py#L1992

genienlp/paraphrase/scripts/transform_dataset.py
Mehrad0711: I added another column to the predictions files to record context as well: https://github.com/stanford-oval/genienlp/blob/93de0965e22fe74257aadad561a7eae2a8bbf26e/genienlp/predict.py#L504

setup.py
Mehrad0711: supporting e2e dialogues will be part of the core library. we can't import this lazily since it will be imported on top of many files.
Optional libraries are those which are purely for human consumption. 
I've kept this as a git dependency for now to track master cause the library is not stable yet and we keep pushing bug fixes etc.
It will become a regular dependency once I do a newer release after I test the whole thing.

genienlp/tasks/generic_task.py
Mehrad0711: Fixed. there were borked during `bitod` --> `e2e_dialogue` renaming.

genienlp/paraphrase/scripts/transform_dataset.py
s-jse: This change to the output breaks a lot of schema2qa scripts in genie-workdirs. They assume the length of each line is 3, and for example use `_id, question, tt = tuple(line.plit('\t'))` in python.
You need to roll this, and the line in predict.py back. If you need this, you should put it under yet another command line argument, until we gradually move to jsonl.

Mehrad0711: hold on, `query file` here is output of genienlp prediction command right? it's not an input.
The format before was `id, pred`.
I added a new column for gold outputs in this commit before: https://github.com/stanford-oval/genienlp/pull/177/commits/3d9708b3cb793e550332fdf8ce9057ba41286e75
And only change needed was ignoring the last entry and it didn't break other scripts afaik. 
Now I'm adding a new column and shifting queries here again. Why would that break anything?

genienlp/metrics.py
```diff
@@ -51,7 +53,8 @@
 
 # metrics that are calculated over a corpus (i.e. a list of predictions and gold answers, not single ones).
 # These metrics cannot be calculated on individual examples and then averaged.
-corpus_level_metrics = set(['bleu', 'casedbleu', 'ter', 't5_bleu', 'nmt_bleu', 'corpus_f1'])
+corpus_level_metrics = {'bleu', 'casedbleu', 'ter', 't5_bleu', 'nmt_bleu', 'corpus_f1'}
```
s-jse: I added this set, but can you double check this is correct?

Mehrad0711: This is a simpler syntax for defining sets :)

s-jse: I know, I was referring to the content of the set, am I missing metrics that are "corpus level" i.e. "cannot be calculated on individual examples and then averaged"?

genienlp/predict.py
Mehrad0711: I know I should have done that but I felt lazy! thanks for not letting me.. I'll make a separate PR for this.

s-jse (approved):
- What caused the expected output for some tasks to change?
- If you can, make `dataset/bitod/` files smaller to make testing faster and debugging easier.

Mehrad0711:
- Adding input column to prediction files. The actual model output remains the same.
- The slowest tests are currently `almond` and `almond_multilingual`: https://app.travis-ci.com/github/stanford-oval/genienlp/builds/247089400   Even fixing those wouldn't help us cause LGTM is currently taking >45 mins.
I suggested removing LGTM before since now we have pre-commit hooks already checking for syntax issues etc. But then saw LGTM still reports some errors not caught during code-check so might as well keep it.

Issue #231, opened 2021-11-30
Support for non-almond datasets


Mehrad0711:
Through several rounds of changes we've made so far, the generic datasets (originally part of decaNLP) cannot be loaded properly in genienlp anymore and should be fixed.
Also, many of those datasets (e.g. iwslt, sst, squad, imdb) are now part of [datasets](https://github.com/huggingface/datasets) library. We should remove those obsolete implementations and associated metrics, and import them from datasets instead.
We have added support to load new datasets from HF datasets. Check out https://github.com/stanford-oval/genienlp/blob/master/genienlp/tasks/hf_dataset.py for the implementation.

sudam802:
may Contribute to this

PR #252, opened 2022-02-28
Major Cleanup #1

Mehrad0711:
A cleanup for genienlp has been long overdue!

We've collected quite some technical debt which we need to address. We have created a [project dashboard](https://github.com/stanford-oval/genienlp/projects/4) to keep track of issues we want to address. I've moved the cards this PR addresses to "In progress".

The goal of this series of cleanups is to simplify the code, make it more readable, easier to modify, and more inviting for outside collaborators.

There are no functional changes (besides dropping almond_multilingual tasks and associated code) - tests are still passing.

We're dropping support for old generic datasets and associated metrics which were inherited from decaNLP code.
Those implementations are obsolete and should be replaced with what is now accessible through [datasets library](https://github.com/huggingface/datasets). Please see [issue](https://github.com/stanford-oval/genienlp/issues/231). If you want support for a new dataset, feel free to submit a PR for it.


genienlp/metrics.py
```diff
@@ -344,184 +173,6 @@ def computeNMTBLEU(outputs, targets):
...
-
-        div_by_zero_error = std_err.decode("utf-8").startswith("Illegal division by zero")
-        if pipes.returncode == 0 or div_by_zero_error:
-            # Still returns the correct output even with div by zero
-            return std_out
-        else:
-            raise ValueError(std_out.decode("utf-8") + "\n" + std_err.decode("utf-8"))
-
-
-def computeROUGE(greedy, answer):
```
s-jse, line 404: Please bring back Rouge, perhaps using HF dataset. It is not just for summarization, it is also used in many natural_seq2seq tasks like paraphrasing. I know it is not on by default, but we do use it from time to time.

s-jse, line 404: Please bring back all Rouge scores (1, 2, L), perhaps using HF dataset. It is not just for the -now deleted- summarization task, it is also used in many `natural_seq2seq` tasks like paraphrasing. I know it is not used by default, but we do use it from time to time.

genienlp/metrics.py
```diff
@@ -604,14 +281,7 @@ def compute_metrics(
...
             sm (structure match): valid if the output is ThingTalk code. Whether the gold answer and prediction are identical if we ignore parameter values of ThingTalk programs
-            bleu
-            rouge1, rouge2, rougeL, avg_rouge
-            f1: token-level F1 score, tokenizes with whitespace
-            nf1: normalize outputs then calculate token-level F1 score
-            nem: normalize outputs then calculate exact match
-            corpus_f1, precision, recall: corpus-level precision, recall and F1 score
-            lfem
-            joint_goal_em, turn_request_em, turn_goal_em, avg_dialogue
+            #TODO add all
```
s-jse: Address the TODO, list what metrics we support.

genienlp/models/base.py
```diff
@@ -86,3 +94,587 @@ def set_generation_output_options(self, tasks):
...
+            outputs = self.model(
+                batch.context.value,
+                labels=batch.answer.value,
+                attention_mask=(batch.context.value != self.numericalizer.pad_id),
+            )
+            return outputs
+        else:
+            return self.model(**kwargs)
+
+    def validate(self, data_iterator, task, original_order=None, disable_progbar=True, **kwargs):
```
s-jse, line 664: This hierarchy of validation/generation functions, their naming and the naming of their outputs are not quite right. For example, when validating a `TransformerSeq2Seq` model, we have the following:

we call `validate()` in `validate.py`
which calls `generate_with_model()`
which calls `validate()` in `models/base.py` which outputs a `GenerationOutput` object
which calls `generate()` in `models/transformer_seq2seq.py`
which calls `generate()` from HF's `transformers` package.

First, there are too many levels, and second, it is really confusing.
You should define how many levels are needed, and move around the functionality until each level does exactly one clearly defined thing. We can discuss this more.

genienlp/models/base.py
```diff
@@ -86,3 +94,587 @@ def set_generation_output_options(self, tasks):
...
+                attention_mask=(batch.context.value != self.numericalizer.pad_id),
+                labels=batch.answer.value,
+            )
+
+            labels = batch.answer.value.tolist()
+
+            logits = output.logits
+            predictions = torch.argmax(logits, dim=-1).tolist()
+
+            # logits for sequence classification is 2 dimensional
```
s-jse, line 690: You have put `TransformerForSequenceClassification` and `TransformerForTokenClassification` under the same parent class.
This comment made me wonder how is it that their validation can have the same implementation and therefore be in the parent class. Don't these models outputs at the very least have different dimensions? One is an `n*c`, the other one is `1*c` (n is sequence length, c is the number of classes)

genienlp/predict.py
```diff
@@ -391,48 +370,71 @@ def prepare_data_iterators(args, val_sets, numericalizer, device):
...
     Model = getattr(models, args.model)
     model, _ = Model.load(
         args.path,
         model_checkpoint_file=args.checkpoint_name,
         args=args,
         device=device,
         tasks=args.tasks,
-        src_lang=src_lang,
-        tgt_lang=tgt_lang,
+        src_lang=args.pred_src_languages[0],
```
s-jse, line 432: So now `pred_src_languages` and `pred_tgt_languages` are only used for determining the language token of pre-trained models that support it?

s-jse (changes requested):
Thanks, overall is a good cleanup, but you should make a few changes before merging it. Please see the individual comments.
Also, have you tested this code with old non-multilingual models? What about the genie-k8s scripts? Especially since you have removed training/prediction arguments.

genienlp/predict.py
Mehrad0711: There are no functional changes. Before you could pass a string of languages (joined by '+') for each task. It was a convenient way of evaluating on multiple languages in one run (e.g. for almond_multilingual) tasks. 
I removed that option since almond_multilingual is gone and the same task can be achieved by running predict multiple times once for each language.

genienlp/models/base.py
Mehrad0711: This exact code does the unification. It turns 2D (batch, class) outputs of sequence classification into 3D (batch, length, class) to conform to 3D output of token classification. 
We then turn each output into a string in the next block of code so everything is flattened out.

genienlp/metrics.py
Mehrad0711: Improving documentation is not of my goals in this PR. Sorry...
I'd rather leave this to future PRs or someone else to address.

genienlp/models/base.py
Mehrad0711: I agree. Lemme see what I can do.

genienlp/train.py
```diff
@@ -228,12 +222,35 @@ def should_log(iteration, log_every):
     return iteration % log_every == 0
 
 
-def do_validate(
-    iteration, args, model, numericalizer, val_iters, *, train_task, round_progress, task_progress, writer, logger
-):
+def validate(task, val_iter, model, args, num_print=10):
```
s-jse: The new changes to the validation code have definitely made it cleaner. Can you change the naming of `validate()` and `do_validate()` in train.py to `validate_during_training()` and `do_validate_during_training()` to further differentiate them from `validate()` defined elsewhere?

genienlp/train.py
```diff
@@ -228,12 +222,35 @@ def should_log(iteration, log_every):
...
+    with torch.no_grad():
+        model.eval()
+        if isinstance(model, torch.nn.DataParallel):
+            # get rid of the DataParallel wrapper
+            model = model.module
+
+        generation_output = model.validate(val_iter, task)
```
s-jse: Can you also rename `generation_output` variable (here and everywhere else) and the `GenerationOutput` class to `validation_output` and `ValidationOutput`? This way, the term "generation" (generating text output) and "validation" (generating text output and then using gold answers to potentially calculate confidence scores or metrics) will have the same meaning everywhere.

genienlp/train.py
Mehrad0711: The only `validate()` function (not attached to a class) we have now is in "train.py". Not sure if we should namespace it. If you're worried about confusion with `validate()` methods in "base.py" I can rename it to `evaluate()` and `do_evaluate()` as that's a shorter and more conventional name (HF use it also). What do you think?

genienlp/train.py
Mehrad0711: Sure

genienlp/train.py
s-jse: Validate and evaluate are often used interchangeably, so doesn't help with confusion. I just think validate_during_training() is more descriptive, and e.g. made me quickly realize I should not use it for the new prediction code I plan to write.

Mehrad0711: Ok let's go with your suggestion

s-jse (approved):

PR #253, opened 2022-03-03
Add evaluate-file command + Some changes

Mehrad0711:
- Add `evaluate-file` command to compute metrics on a file. The file format should be the same as the output file of the `predict` command.
- Add the option to start validating the model after a certain number of training iterations have passed.
- Update some tests.


genienlp/evaluate_file.py
```diff
@@ -0,0 +1,188 @@
...
+    parser.add_argument(
+        '--pred_tgt_languages',
+        type=str,
+        nargs='+',
+        default=['en'],
+        help='Specify dataset target languages used during prediction for multilingual tasks',
+    )
+
+    parser.add_argument(
+        '--main_metric_only', action='store_true', help='If True, we only calculate the deca score metric for each task.'
```
s-jse, line 85: Add a check that raises `ValueError` if `main_metric_only` and `extra_metrics` are both specified. The same in `predict.py`.

genienlp/models/base.py
```diff
@@ -409,6 +409,11 @@ def validate_e2e_dialogues(
         device = self.device
         args = self.args
 
+        if self.numericalizer._tokenizer.src_lang:
+            src_lang = self.numericalizer._tokenizer.src_lang
+        else:
+            src_lang = self.orig_src_lang
```
s-jse, line 415: It seems to me that `self.orig_src_lang` can be `None` in this line for HF tokenizers that don't have language codes. Is `src_lang` part of their API and therefore guaranteed for all tokenizers? If not, wouldn't it break `lang=src_lang[:2]` in line 534, or similarly in line 194?

setup.py
```diff
@@ -48,7 +48,7 @@
     install_requires=[
         'numpy>=1.14.5',
         'torch>=1.9.0,<1.11.0',
-        'pyrouge>=0.1.3',
+        'rouge_score>=0.0.4',
```
s-jse, line 51: Doesn't HF dataset already install this as a requirement?

s-jse (approved):

setup.py
Mehrad0711: Nope. Same as bert-score we need to install dependencies separately.

genienlp/models/base.py
Mehrad0711: we set `src_lang` and `tgt_lang` attribute for all the tokenizers here: https://github.com/stanford-oval/genienlp/blob/fb35befc2ed2a1c606b9b1dbbb7c3c5df1a46b7e/genienlp/data_utils/numericalizer.py#L216
`src_lang` is the adjusted version of `orig_src_lang`: https://github.com/stanford-oval/genienlp/blob/fb35befc2ed2a1c606b9b1dbbb7c3c5df1a46b7e/genienlp/util.py#L740
`orig_src_lang` is only used for Marian tokenizers. For single-language Marian models (those that source and target language are only one language) we set `src_lang` and `tgt_lang` to None since they don't need it. Check this code for more info: https://github.com/stanford-oval/genienlp/blob/fb35befc2ed2a1c606b9b1dbbb7c3c5df1a46b7e/genienlp/models/transformer_seq2seq.py#L60

PR #263, opened 2022-05-09
Wip/dialogues

Mehrad0711:
Main changes:
- A two-step translation approach where we first build a dictionary of possible translations of an entity. We then use that for alignment after sentence translation. If translation of the entity is not in the dictionary we use the neural alignment.
- Misc. fixes to e2e_dialogues code: fixing JGA evaluation, cleaning up old code, etc.

### DIFF
--- a/genienlp/predict.py
+++ b/genienlp/predict.py
@@ -246,6 +246,11 @@ def parse_argv(parser):
         action='store_true',
         help='do not preserve quotation marks in the output. Useful if using alignment for semantic parsing or NLG',
     )
+    parser.add_argument(
+        '--align_span_symbol',
+        type=str,
+        help='The symbol we use to wrap spans of words in the input that need to be preserved in the output.',
+    )
 
     parser.add_argument(
         '--e2e_dialogue_evaluation',
--- a/genienlp/predict.py
+++ b/genienlp/predict.py
@@ -387,8 +387,12 @@ def prepare_data_iterators(args, val_sets, numericalizer, device):
     return iters
 
 
-def create_output_lines(args, index, validation_output):
-    predictions = validation_output.raw_predictions if args.translate_return_raw_outputs else validation_output.predictions
+def create_output_lines(args, index, validation_output, raw_outputs=False):
+    if raw_outputs and args.translate_return_raw_outputs:
+        predictions = validation_output.raw_predictions
+    else:
+        predictions = validation_output.predictions
+
     if args.one_output_per_line:
         lines = [
             '\t'.join(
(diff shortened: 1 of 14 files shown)


genienlp/train.py
```diff
@@ -516,16 +516,25 @@ def train(
...
+                        model,
+                        batch,
+                        iteration,
+                        opt,
+                        devices,
+                        lr_scheduler=lr_scheduler,
+                        grad_clip=args.grad_clip,
+                        gradient_accumulation_steps=args.gradient_accumulation_steps,
+                    )
+                except RuntimeError as e:
```
s-jse, line 530: Why? OOM should not be demoted to a warning.

genienlp/tasks/almond_task.py
```diff
@@ -503,6 +548,10 @@ def batch_postprocess_prediction_ids(self, batch_example_ids, batch_src_ids, bat
 
         return partial_batch_prediction_ids, all_text_outputs
 
+    # def get_splits(self, root, **kwargs):
```
s-jse: Please remove commented lines if they are not needed.

genienlp/predict.py
```diff
@@ -223,6 +223,11 @@ def parse_argv(parser):
         action='store_true',
         help='split examples with multiple sentences into individual examples',
     )
+    parser.add_argument(
+        '--translate_only_entities',
```
s-jse, line 227: **only** entities as opposed to what? translating full sentences? Can you explain what this does?
We might need to use a more descriptive name instead of `translate_only_entities`.

s-jse (approved):
Thanks. I added a few comments.

genienlp/train.py
Mehrad0711: Since OOM is caused by hardware limits, we should let users set their tolerance for it. I've noticed OOM even with small batch sizes in some of my experiments on bitod. I can make it a cmd-line argument e.g. `--allow_OOM`

genienlp/predict.py
Mehrad0711: Yeah as opposed to the full sentence. If this argument is provided, the entities are extracted from the sentence and translated individually.  The output will be a dictionary of entities and their translations per sentence.

genienlp/train.py
s-jse: But this has the potential to give poor results during training, as only part of the training data is seen. Validation results might also be different on different hardware, without users noticing.
I think it is reasonable to assume that if people cannot fit even a single example of a dataset on their GPU, they should find another way to run that code. Do we have an issue like that where even single examples don't fit?

Mehrad0711: > But this has the potential to give poor results during training, as only part of the training data is seen.

Not necessarily. I have run experiments before and the results are good. Usually, only a few examples are problematic and will be skipped.

> Validation results might also be different on different hardware, without users noticing.

Yeah, this is why I think making it optional is reasonable with a cmd-line argument that is off by default.

> I think it is reasonable to assume that if people cannot fit even a single example of a dataset on their GPU, they should find another way to run that code. Do we have an issue like that where even single examples don't fit?

Yeah if you're trying very big models there's no point in using this option. The option is useful when you get very few OOMs e.g. once or twice per epoch.

PR #270, opened 2022-06-16
Wip/dialogues

Mehrad0711:
Main changes:
- Add SER, DST_em, and DA_em metrics
- Some refactoring to avoid curcular import errros when importing genienlp in genie-workdirs scripts
- Minor code cleanups

### DIFF
--- a/genienlp/arguments.py
+++ b/genienlp/arguments.py
@@ -564,7 +564,7 @@ def parse_argv(parser):
         '--e2e_dialogue_valid_submetrics',
         nargs='+',
         type=str,
-        default=['jga', 'em', 'em'],
+        default=['em', 'em', 'em'],
         help='Specify metrics to use for each of subtasks in e2e_dialogue_valid_subtasks.',
     )
     parser.add_argument(
--- a/genienlp/arguments.py
+++ b/genienlp/arguments.py
@@ -564,7 +564,7 @@ def parse_argv(parser):
         '--e2e_dialogue_valid_submetrics',
         nargs='+',
         type=str,
-        default=['em', 'em', 'em'],
+        default=['dst_em', 'em', 'da_em'],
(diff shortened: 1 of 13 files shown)


genienlp/arguments.py
```diff
@@ -564,7 +564,7 @@ def parse_argv(parser):
         '--e2e_dialogue_valid_submetrics',
         nargs='+',
         type=str,
-        default=['jga', 'em', 'em'],
+        default=['dst_em', 'em', 'da_em'],
```
s-jse, line 567: To confirm I understand this correctly, `em` is actual exact match, `dst_em` and `da_em` normalize first, then do exact match, right? Can you mention this in a comment?

genienlp/metrics.py
```diff
@@ -179,25 +179,31 @@ def computeNMTBLEU(outputs, targets):
     return bleu_metric.compute(predictions=outputs, references=targets)['bleu'] * 100
 
 
-def compute_e2e_dialogue_score(greedy, answer, tgt_lang, args, example_ids):
+def compute_e2e_dialogue_score(greedy, answer, tgt_lang, args, example_ids, contexts):
```
s-jse, line 182: I think we should use `predictions` and `answers` (plural) instead of `greedy` and `answer` (like in `compute_ner_f1`). Makes it less confusing if someone is not using greedy decoding.

s-jse (approved):
Looks good overall, I left some minor comments.

genienlp/arguments.py
Mehrad0711: Yeah

genienlp/metrics.py
Mehrad0711: Yeah, we need to update all function signatures in metrics to be consistent. Right now we have `(outputs, targets)`, `(greedy, answer)`, or `(predictions, answers)`. I added it to the cleanup TODOs: https://github.com/stanford-oval/genienlp/projects/4

PR #276, opened 2022-06-29
Add dialogue loop + risawoz task

Mehrad0711:
- Add a dialogue loop so users can interact with the dialogue agent (over cmdline)
- Add RiSAWOZ task
- Update dialogues-related code to agree with https://github.com/stanford-oval/dialogues/pull/5


genienlp/metrics.py
```diff
@@ -224,23 +223,38 @@ def compute_e2e_dialogue_score(greedy, answer, tgt_lang, args, example_ids, cont
...
     act_values = []
     for input in inputs:
         act_values.append(QUOTED_MATCH_REGEX.findall(input))
 
-    return compute_ser(greedy, act_values)
+    return dataset.compute_ser(greedy, act_values)
 
 
-def computeDA_EM(greedy, answer):
+def computeDA_EM(greedy, answer, task_names):
```
s-jse, line 239: what is `task_names`? what is its purpose?

genienlp/models/base.py
```diff
@@ -125,6 +125,24 @@ def __init__(
...
 # TransformerSeq2Seq and TransformerLSTM will inherit from this model
 class GenieModelForGeneration(GenieModel):
+    def numericalize_example(self, input_text, turn_id, device):
+        if isinstance(input_text, str):
+            input_text = [input_text]
+        tokenized_contexts = self.numericalizer.encode_batch(input_text, field_name='context', features=None)[0]
+
+        numericalized_turn = NumericalizedExamples(
+            example_id=[str(turn_id)],
+            context=SequentialField(
```
s-jse, line 135: Is there a reason we don't just use `tokenized_contexts` instead of creating a new `SequentialField`?

genienlp/tasks/generic_task.py
```diff
@@ -253,8 +253,65 @@ def get_splits(self, root, **kwargs):
...
+    def get_splits(self, root, **kwargs):
+        kwargs['e2e_evaluation'] = self.args.e2e_dialogue_evaluation
+        return E2EDialogueDataset.return_splits(path=root, make_example=self._make_example, **kwargs)
+
+
+@register_task('risawoz')
+class RiSAWOZ(E2EDialogueTask):
+    def __init__(self, name, args):
+        super().__init__(name, args)
+        self.special_tokens = {
```
s-jse: I think we have two different `special_tokens` in GenieNLP:

1. The `--preprocess_special_tokens` argument in `arguments.py` and related code which means we replace long ThingTalk tokens with a smaller prefix as a preprocessing step.
1. The `special_tokens` here, which are supposed to be treated as one token by the tokenizer?

Am I understanding this correctly? If so, why use the same name?

genienlp/tasks/generic_task.py
```diff
@@ -253,8 +253,65 @@ def get_splits(self, root, **kwargs):
         return OODDataset.splits(root=root, **kwargs)
 
 
+class E2EDialogueTask(BaseTask):
```
s-jse: I think this is time to move dialogue tasks and datasets to separate files named `dialogue_task.py` and `dialogue_dataset.py`

s-jse (approved):
Looks good to me. I added some comments below.

genienlp/metrics.py
Mehrad0711: `task_names` is a list of tasks you want to evaluate - for example to evaluate a model trained on multiple tasks simultaneously.
We pass it here to be able to initialize the correct dataset (e.g. bitod, risawoz, etc.) from dialogues.

genienlp/tasks/generic_task.py
Mehrad0711: Yeah `special_tokens` name is a bit overloaded in the code, but we are essentially using it for any token that should not be touched by the tokenizer. `--preprocess_special_tokens` should really be `--preprocess_thingtalk_tokens`. What it does is breaking thingtalk tokens into smaller tokens that are then treated as special tokens.

genienlp/models/base.py
Mehrad0711: `tokenized_contexts` attributes have list type. We need to turn them into tensors before passing it to the model.

genienlp/tasks/generic_task.py
Mehrad0711: Sure

genienlp/metrics.py
s-jse: This way (`task_names[0].split('_', 1)[0].capitalize()`) of going from task name to dataset class can cause bugs down the line, if you change any task or dataset names. Please add a mapping somewhere in dialogue_task.py instead.

Mehrad0711: Ok. I added it to the branch. Will PR with other changes later.

PR #280, opened 2022-07-15
Update e2e dialogues

Mehrad0711:


genienlp/arguments.py
```diff
@@ -126,6 +126,9 @@ def parse_argv(parser):
     parser.add_argument(
         '--num_print', default=10, type=int, help='how many validation examples with greedy output to print to std out'
     )
+    parser.add_argument(
+        '--topk_print', default=3, type=int, help='print length of the top K longest examples as well as min, max, and mean'
```
s-jse, line 130: "top k" in this repository is used to refer to top generated tokens at each generation output; we even have another argument called `--top_k`.
Please either change the name, or remove it and make it always print the 3 longest examples, if 3 is always enough for you application

Mehrad0711: I knew you were going to say this!
I need it to be configurable. Do you have a suggestion for the name (preferably short)?

s-jse: Something like `log_n_examples` or `log_n_example_lengths`?

PR #285, opened 2022-08-05
Misc. updates

Mehrad0711:
- Fix and update the implementation of sentence classification task; add error detection task on top of it
- Filter out examples longer than the model maximum input length
- Some other cosmetic changes


genienlp/models/base.py
```diff
@@ -68,15 +68,6 @@ def load(cls, save_directory: str, *model_args, **kwargs):
         model = cls(args=args, tasks=tasks, vocab_sets=vocab_sets, save_directory=save_directory, *model_args, **kwargs)
         save_dict = torch.load(full_checkpoint_path, map_location=device)
 
-        # HACK
```
s-jse: Why did you remove this? We still use some older models that are trained with `transformers==4.0`, like the paraphraser.

genienlp/tasks/dialogue_dataset.py
```diff
@@ -47,3 +47,7 @@ def return_splits(cls, path='.data', train='train', validation='valid', test='te
         return Split(train=train_data, eval=validation_data, test=test_data), Split(
             train=train_path, eval=validation_path, test=test_path
         )
+
+
+class ErrorClassificationDataset(E2EDialogueDataset):
```
s-jse: If this class is specific to dialogue datasets (which I think it is, given that it inherits from `E2EDialogueDataset`, it should have "Dialogue" in its name.

genienlp/tasks/dialogue_task.py
```diff
@@ -131,3 +131,53 @@ def get_splits(self, root, **kwargs):
         kwargs['train_target'] = 'dst'
         kwargs['e2e_evaluation'] = self.args.e2e_dialogue_evaluation
         return E2EDialogueDataset.return_splits(path=root, make_example=self._make_example, **kwargs)
+
+
+@register_task('error_cls')
+class ErrorClassificationTask(BiTOD):
```
s-jse: A similar problem here. If this is BiTOD-specific, it should have BiTOD in the name, if not, make sure the code is generic enough that works with all dialogue datasets.

s-jse (changes requested):
A few comments below. Can you please also add a test for the new sentence classification task?

genienlp/models/base.py
Mehrad0711: This hack caused dimension errors when loading a trained bart model for sequence classification (with the newest version of transformers). I think best solution is to retrain the paraphraser with the new version if it's only a few models. Otherwise we need to update the condition here as to not interfere with models with a classification head on top. But hacks are generally not good and may cause silent errors again down the line. Finding this was causing the error already took some time...

s-jse (approved):

PR #290, opened 2022-09-02
Change batching algorithm

s-jse:
- Change batching algorithm
**This change does not affect the quality of trained models, but will make them not directly comparable to previous versions due to different randomness.**
Before this change, we would sort the training data by their size (e.g. number of tokens in the input and output) and then sample from this sorted list by picking a starting index and adding subsequent examples to the batch until the batch is full. This was done so that we can group examples of similar size into the same batch, and decrease wasted GPU computations by decreasing the number of needed pad tokens.
This had two downsides:
1.  It was not guaranteed that we will see every training example the same number of times during training. While acceptable in theory and especially large synthetic datasets, not seeing every training example is not desirable in general.
1.  This sampling algorithm actually favors smaller examples, since batches containing smaller examples contain more examples.

In this PR, we fix these two, while maintaining the speedup from batching examples of similar size together. We see every training example exactly `n` times before we see any example `n+1` times. I also tested that the training speed stays the same after this change.
This is done by "marking" seen examples in each epoch, and only sampling from the unmarked examples. Marks get cleared after the epoch ends.

- Add documentation for optional dependencies
- Add progress bar for when the validation dataset is large

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,13 @@ pip install -e .  # -e means your changes to the code will automatically take ef
 
 After installation, `genienlp` command becomes available.
 
+Some GenieNLP commands have additional dependencies for plotting and entity detection. If you are using those commands, you can obtain their dependencies by running the following:
+
+```
+pip install matplotlib~=3.0 seaborn~=0.9
+python -m spacy download en_core_web_sm
+```
+
 ## Usage
 
 ### Training a semantic parser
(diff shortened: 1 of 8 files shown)


s-jse:
@Mehrad0711 , do you think we should hide the new batching algorithm behind a flag, or is it good this way? I ran a training job on Schema2QA/restaurants before and after this change. Green line is `master`, red line is this PR:
image

Mehrad0711:
I think we need to do some more testing esp. on other tasks, to ensure there is no performance degradation.
I prefer adding a flag too - you can make the new behavior default. We will drop it later if everything is fine.

Mehrad0711 (approved):

